PR #41, opened 2015-09-04
fixes #39 require either title or bibId

### LINKED ISSUE #39: When ingesting a book you should be able to supply a title OR a bibID

flyingzumwalt:
... and don't attempt to retrieve external metadata if bibId isn't present


flyingzumwalt:
fixes #39
fixes #33

Issue #50, opened 2015-09-05
IIIF predicates on books and pages


jpstroop:
Somewhat related to #32 in that these must be editable fields, though not descriptive properties. Will be required for #21.

(`sc` = `http://iiif.io/api/presentation/2#`)

__Scanned Books need:__

Viewing Direction (`sc:viewingDirection`). The value/object must be one of:
 * left-to-right (`sc:leftToRightDirection`) __(default)__
 * right-to-left (`sc:rightToLeftDirection`)
 * top-to-bottom (`sc:toToBottomDirection`)
 * bottom-to-top (`sc:bottomToTopDirection`)

Viewing Hint (`sc:viewingHint`). The value/object must be one of:
 * individuals (`sc:individualsHint`) __(default)__
 * paged (`sc:pagedHint`)
 * continuous (`sc:continuousHint`)

__Pages need:__

Viewing Hint (`sc:viewingHint`). The value/object must be one of:
 * non-paged (`sc:nonPagedHint`)
 * facing-pages (`sc:facingPagesHint`)

Label (`rdfs:label`)

terrellt:
All dropdowns.

terrellt:
@jpstroop I don't see the URIs for the acceptable values (IE `sc:leftToRightDirection`) in the spec, so I assume they won't resolve. Rather I just see literals as stand-ins - is this work that's in progress?

jpstroop:
Hmm. The URIs are used in the the [JSON-LD context](http://iiif.io/api/presentation/2/context.json), and I thought that they did resolve to [this ontology](http://iiif.io/api/presentation/2/ontology.xml).

Paging @azaroth42: Any reason we can't make this so?

terrellt:
Oh, that's interesting. The JSON-LD context layers a URI on the literal

azaroth42:
While URIs SHOULD resolve to their appropriate term definitions, what's the requirement in particular?

`left-to-right` is `sc:leftToRightDirection` is `http://iiif.io/api/presentation/2#leftToRightDirection`
So it's a URI ... which doesn't resolve.

I put in a redirect from 2 to 2/ontology.xml on the iiif.io server and created a "Update the ontology" issue:  https://github.com/IIIF/iiif.io/issues/551

terrellt:
@azaroth42 :+1: That'll fix resolution. Thank you.

[...]

PR #53, opened 2015-09-08
Add IIIF schema.

terrellt:
Work towards #50 

What this doesn't include is validation and any form work.


jpstroop:
Should we consider adding it to the RDF::Vocab gem once IIIF/iiif.io#551 is fixed?

terrellt:
@jpstroop https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/iiif.rb The ontology is there already, this is more like the accessors for mapping from instance methods to predicates.

terrellt:
I added a spec in there to make sure the viewing hint on GenericFile would update, and it appears to - maybe I should just delete that spec?

jpstroop:
I guess you're technically testing the behavior of GenericFile here and therefore shouldn't have to, but I'm not going to complain about an extra spec.

terrellt:
@jpstroop I'm gonna remove it and trust our layers. Extra second of spec for something I can't make fail doesn't seem worth it.

jpstroop:
Heh, fair enough. But did we decide that we're storing the literals instead?

https://github.com/pulibrary/plum/blob/iiif_properties/spec/models/generic_file_spec.rb#L8

jpstroop:
(Maybe I don't understand what that code is doing well enough....)

terrellt:
@jpstroop That's just making sure that the method #viewing_hint maps to the viewingHint predicate. The literals stuff will come in for the validation PR.

PR #54, opened 2015-09-08
Update pul_metadata_services.

### LINKED ISSUE #1: Upgrade to the latest version of pul_metadata_services
The API for pul_metadata_services has changed but pul-worthwhile was not updated to consume it.  For now, I've pinned to the old version of pul_metadata_services.  Update the gemfile to use the latest version and refactor the code as necessary.

terrellt:
Closes #1 

If we merge this we'll need another ticket to unmark the EAD spec as pending.


jpstroop:
Had this in progress before I had to cut out. Yours is prettier.

app/models/scanned_book.rb
```diff
@@ -5,6 +5,7 @@ class ScannedBook < ActiveFedora::Base
   include ::CurationConcerns::BasicMetadata
   include ::NoidBehaviors
 
+  property :sort_title, predicate: ::RDF::URI.new("http://library.princeton.edu/terms/primary_title"), multiple: false
```
terrellt: We might want to find a real predicate for this.

jpstroop:
I'm :ok: with merging, but can't from my phone for some reason.

app/models/scanned_book.rb
jpstroop: I feel like I've looked for one before without luck or range/domain restrictions, but it's been a while.

It's one of those things that's far too practical for linked data :wink:

terrellt:
SKOS.prefLabel gets close, but may not have the semantics of "this is a title'

terrellt:
@jpstroop I've got a branch that puts sort_title in DCT.title and title in DCT.alternative. Is that accurate?

jpstroop:
Other way around, I think. And where do all the other non-main entry titles
go (MARC records can have quite a few)? I still think we can need a
subproperty of dct:alternative that holds exactly one value which users do
not generally see but allows the UI to sort objects by title properly.

Nobody ever sees this title. It's just there so that

-Js

terrellt:
Oh, right, I had this conversation with @scande3 a while back. A title to sort on because for some reason solr's processors don't do well enough. https://t.co/RMlXyXKq6W seems relevant (from @scande3)

jpstroop:
Yup, exactly.

jpstroop:
`mods:nonSort` and MARC 245 second indicator (which is informing what we're doing here) are functionally the same.

terrellt:
@jpstroop Switched to using that group's recommended predicate (which doesn't exist yet, but it could!)

PR #55, opened 2015-09-08
Make title readonly if there's a source ID.

### LINKED ISSUE #46: Title (and other metadata) should be made read only if it's set from remote metadata

terrellt:
Closes #46


app/views/curation_concerns/scanned_books/_form_required_information.html.erb
```diff
@@ -1,5 +1,6 @@
 <fieldset class="required">
   <legend>Required Information</legend>
+  <%= f.input :title, readonly: f.object.source_metadata_identifier.present? %>
```
terrellt: It seems like we don't have form objects in Curation Concern forms, which is troublesome for things like this - that'd be a nice clear place to define #readonly?(:title)

PR #65, opened 2015-09-10
Jp2 derivatives

### LINKED ISSUE #26: Generate JP2s on ingest

terrellt:
Closes #26

### DIFF
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -4,4 +4,19 @@ class GenericFile < ActiveFedora::Base
   apply_schema IIIFPageSchema
 
   validates_with ViewingHintValidator
+  makes_derivatives :create_intermediate_file
+
+  private
+
+    def create_intermediate_file
+      case original_file.mime_type
+      when 'image/tiff'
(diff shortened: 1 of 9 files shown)


terrellt:
Considered indexing, but usage should probably be dealt with based upon how to access Loris.

app/models/generic_file.rb
```diff
@@ -4,4 +4,19 @@ class GenericFile < ActiveFedora::Base
   apply_schema IIIFPageSchema
 
   validates_with ViewingHintValidator
+  makes_derivatives :create_access_copy
+
+  private
+
+    def create_access_copy
```
jpstroop: It's a small thing, but if we were putting this in the repo, I think we'd be calling this an ["Intermediate File"](https://github.com/duraspace/pcdm/blob/master/pcdm-ext/use.rdf#L28). Maybe rename this method to `:create_intermediate_file`?  ( I realize it's private, but the `makes_derivatives` call near the top caught my eye).

terrellt: @jpstroop Renamed

PR #68, opened 2015-09-11
IIF Manifests

### LINKED ISSUE #21: There should be a IIIF manifest for the book.
Review the [IIIF Presentation API 2.1 Draft](http://prezi-21.iiif.io/api/presentation/2.1/) before taking this on.

Briefly, books that are a single volume will be Manifests, and multi-volume sets will be Collections with viewingHint set to "multi-part".

Also, [o'sullivan](https://rubygems.org/gems/osullivan) is your friend here.

terrellt:
This allows you to add `/manifest` to the end of an item record page to
get its IIIF Presentation Manifest. Requires that Loris is running in
order to get the height/width of the JP2 derivatives.

Closes #21


terrellt:
A note - this may be slow. May want to do some work to make it not depend on Fedora at all

jpstroop:
+1--there's nothing in there manifest that requires Fedora, right? I wonder if it's worth even going a step further and caching the manifest??

terrellt:
@jpstroop I'm not sure it's worth figuring out cache invalidation given the speed of solr

jpstroop:
I was thinking more about having to hit t he image server once per page whenever the manifest is requested. We could index the width and height and not use `IIIF::Presentation::ImageResource#create_image_api_image_resource`, I suppose.

jpstroop:
Looks fine to me--I can make a separate issue re :arrow_up:. Are you ready to have it merged?

terrellt:
@jpstroop Wanted to get it running from solr, since Fedora is basically a non-starter - it would have failed us the moment we put in a real book.

terrellt:
@jpstroop @flyingzumwalt Ready for review

spec/factories/scanned_books.rb
```diff
@@ -4,6 +4,7 @@
     source_metadata_identifier "1234567"
     access_policy "Do you want a jelly baby?"
     use_and_reproduction "Jamie, remind me to give you a lesson in tying knots, sometime."
+    description "900 years of time and space, and I’ve never been slapped by someone’s mother."
```
jpstroop: :+1:

jpstroop:
No complaints here.

PR #69, opened 2015-09-11
Pairtree Derivative Path override.

### LINKED ISSUE #66: Put derivatives in buckets
* Given the pid `df69azpl`
 * The jp2 should be stored at: `{resolver_root}/df/69/az/pl.jp2` (on the mounted image server storage)
 * So that we can use `http://{server}/loris/df%2F69%2Faz%2Fpl.jp2/...` in views to make image requests

terrellt:
Closes #66


terrellt:
Closing - looking like we're pushing this into Curation Concerns.

jpstroop:
+1, I was thinking yesterday that this would be a good idea. Just make sure that the method that calculates the relative path is public so that we can also use it to build the Loris idenifier.

terrellt:
Re-opened as #70

Issue #64, opened 2015-09-10
PDF Generator uses IIIF to properly size page images


flyingzumwalt:
Once IIIF services are available, refactor GeneratesPdfs#to_pdf to use those services.  It should resize the image to fit the page and render it as either a PNG or JPEG.  

Note: You must make sure that images don't exceed the page size.  Default page size is LETTER (612.00 x 792.00).  Currently I'm just telling prawn to make pages with no margins and to limit image height to 792.00.  Ideally this should be replaced by code that uses IIIF to make images that are as close to the page dimensions as possible without exceeding them.

tpendragon:
Duplicate of #76

Issue #25, opened 2015-09-03
As a user I should be able to ask for a PDF representation of the book (rights permitted)


tpendragon:

flyingzumwalt:
Questions (@jpstroop):

* Where should generated pdfs be stored?
* What should the filenames of generated pdfs be set to?
* What metadata should be embedded in the pdf?  Maybe a URI for the scanned_book?

There's a first pass at this functionality in the pdf branch.  Here's what it does: 
* `scanned_book.to_pdf` generates a Prawn::Document with a page for each generic_file in the book
* Each page in the pdf containins the original_file from the corresponding generic_file object
* Most of the scanned_book metadata gets dumped into the PDF's EXIF
* `scanned_book.render_pdf(file_path)` renders the pdf file 

Notes: (see #64)
* I'm currently just grabbing the .original_file.content from each generic_file.  This should be changed once JP2 support is set up
* you must make sure that images don't exceed the page size.  Default page size is LETTER (612.00 x 792.00).  Currently I'm just telling prawn to limit image height to 792.00.

tpendragon:
@jpstroop Can we close this? Maybe open an issue for adding metadata to the PDF?

jpstroop:
:+1:

PR #71, opened 2015-09-12
Adds ability to generate and download PDF from ScannedBook

flyingzumwalt:
ref #25

Does not add a link/button to show view, but you can trigger this action by visiting /concerns/scanned_books/:id/pdf

Note: There's still work to do, recorded in #64


jpstroop:
@terrellt can you review when you get a chance? We should probably merge and refine in other issues, but here are a few thoughts:
 1. Should this be made at the same time as other derivatives, or on the first request and then cached with the derivatives? (I believe we're assuming the latter right now).
 2. It should probably respect the IIIF non-paged `viewingHint`, and perhaps ultimately be derived from the Manifest in general. This would allow us to have an [outline](http://prawnpdf.org/api-docs/2.0/Prawn/Outline.html) as well, which would be pretty snazzy.

terrellt:
@jpstroop
1. My inclination is on first request.
2. Agree. #76.

terrellt:
I'm inclined to merge this and push further work down via #76 and #64. Thoughts @jpstroop?

jpstroop:
:+1: As discussed, I'll make a PDF meta ticket that logs some things to consider as we refine this.

PR #78, opened 2015-09-15
Add branding. Closes #38

### LINKED ISSUE #38: Branding

axapu:


terrellt:
@axapu This looks awesome! The only thing that stands out is when you're logged in the buttons on the top right stack on top of each other

axapu:
@terrellt I'm having trouble getting the other buttons to show up. Any pointers? I'm going to refer to the curation_concerns gem in the meantime and resolve the fix to the button stacking issue.

My view:
screen shot 2015-09-15 at 7 24 06 pm

terrellt:
@axapu Ah, you'll have to make yourself an admin in the rails console. Something like

```
u = User.last
u.roles << Role.create(name: "admin")
```

terrellt:
My view:

screen shot 2015-09-15 at 4 28 31 pm

axapu:
@terrellt Made some changes. Should be better now.

terrellt:
@axapu Yup, looks better! The last thing I see is the buttons fall below the banner coloring in mobile sizes.

axapu:
@terrellt Added some changes to for smaller viewports. Let me know what you think.

terrellt:
Looks great @axapu! This may be a ticket for later, but what's your opinion on overriding bootstrap's variables for styling, rather than adjusting the CSS directly? So replacing things like https://github.com/pulibrary/plum/pull/78/files#diff-1859ff4dcfb99e5b9f6b494236a49aa0R127 with $navbar-default-bg: $light-gray; (from https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_navbar.scss#L398). Similar things for header margins/line-heights as well (https://github.com/twbs/bootstrap-sass/blob/master/templates/project/_bootstrap-variables.sass#L65)

axapu:
We want to eventually cut ties with bootstrap, but all the rails apps we've been working on have it as a dependency. Good call on overriding variables, @terrellt. I got rid of the require tree so we can do this. We can look into this further and dry it up even more in the near future.

PR #84, opened 2015-09-28
Add a basic IIIF viewer.

terrellt:


kevinreiss:
:+1: Definitely a fan of data attributes. @jpstroop this seems a good interim solution until we get a chance to engage with the Universal Viewer more and see if it will place nice with the rails asset pipeline.

jpstroop:
:+1:

app/presenters/scanned_book_show_presenter.rb
```diff
@@ -3,8 +3,11 @@ class ScannedBookShowPresenter < CurationConcerns::GenericWorkShowPresenter
 
   def file_presenters
     @generic_files ||= begin
```
jpstroop: @terrellt Just curious...I've seen you do this a few times. What's the advantage of `begin` / `end` when there's no `rescue` or `ensure`?

terrellt: Oh, I usually do it in this case so that the @generic_files ||= is scoped to multiple lines, so none of it will run if the instance variable is cached. Equivalent would be...

```ruby
def file_presenters
  return @generic_files if @generic_files
  ids = solr_document.fetch('generic_file_ids_ssim', [])
  @generic_files = CurationConcerns::PresenterFactory.build_presenters(generic_file_ids, GenericFilePresenter, current_ability)
end
```

In this refactor I could have removed it entirely - missed that fact.

jpstroop: Got it. Makes sense. :+1:

PR #85, opened 2015-09-30
Add sidekiq as production ActiveJob

terrellt:


config/deploy.rb
```diff
@@ -45,3 +45,19 @@
     end
   end
 end
+namespace :sidekiq do
+  task :quiet do
+    on roles(:app) do
+      # Horrible hack to get PID without having to use terrible PID files
+      puts capture("kill -USR1 $(sudo initctl status plum-workers | grep /running | awk '{print $NF}') || :")
```
jpstroop: Does this work? Wondering about `sudo` ?

terrellt: Yeah. Had to allow the deploy user to use passwordless sudo with these two commands. These are the only two commands it can sudo though.

jpstroop: Yuck. :ok_hand: Having to remember to set that up when we deploy to a new box is going to be a pain

PR #114, opened 2015-10-20
Hiding add content section - fixes #99

### LINKED ISSUE #99: Remove Add Your Content from Create Scanned Book

escowles:


jpstroop:
It never occurred to me that the best way to keep a partial from rendering was to override it with an empty file. Seems like a good approach--are there any drawbacks?

tpendragon:
@jpstroop Just whatever processing power Rails needs to render a partial.

jpstroop:
:ok_hand:

PR #117, opened 2015-10-20
Excluding /spec from Coveralls, using vanilla SimpleCov locally

escowles:
Any opinions on running coverage locally @tpendragon @jpstroop?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ noid-minter-state
 
 /jetty
 /fits
+/coverage
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,7 +1,12 @@
+require 'simplecov'
 if ENV['CI']
   require 'coveralls'
-  Coveralls.wear!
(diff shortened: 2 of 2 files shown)


tpendragon:
Shouldn't we be using SimpleCov.start 'rails'?

jpstroop:
@tpendragon ?

jpstroop:
@tpendragon, Oops, your comment showed up after refreshing. I don't have an opinion. [SimpleCov.start 'rails'](https://github.com/colszowka/simplecov#getting-started) does seem like a thing, however, it may only apply...

> when testing a server process (i.e. a JSON API endpoint) via a separate test process (i.e. when using Selenium) where you want to see all code executed by the rails server, and not just code executed in your actual test files, you'll want to add something like this to the top of script/rails (or bin/rails for Rails 4)

tpendragon:
@jpstroop @escowles `SimpleCov.start 'rails'` has always just meant "start up simplecov with config options that make sense for rails projects" to me. So it ignores /spec and categorizes things appropriately.

PR #118, opened 2015-10-20
Displaying error message when source metadata can't be retrieved - closes #98

### LINKED ISSUE #98: If I submit a new record with a source metadata ID it can't get data for it should error out

escowles:


spec/controllers/curation_concerns/scanned_resources_controller_spec.rb
```diff
@@ -24,6 +24,18 @@
         expect(s.title).to eq ['The Giant Bible of Mainz; 500th anniversary, April fourth, fourteen fifty-two, April fourth, nineteen fifty-two.']
       end
     end
+    context "when given a non-existent bib id", vcr: { cassette_name: 'bibdata_not_found', allow_playback_repeats: true } do
+      let(:scanned_resource_attributes) do
+        FactoryGirl.attributes_for(:scanned_resource).merge(
+          source_metadata_identifier: "0000000"
+        )
+      end
+      it "receives an error" do
```
tpendragon: Do we need a statement to make sure a ScannedResource doesn't get created?

escowles: Good idea -- PR updated!

PR #121, opened 2015-10-21
Update Curation Concerns

tpendragon:
This updates to the latest curation concerns, which adds new ordering
from ActiveFedora::Aggregations and the new interface for
Hydra::Derivatives.


spec/factories/generic_files.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
```
escowles: Do you want to rename the file to file_sets.rb to keep it in sync with the factory name?

spec/support/features/session_helpers.rb
```diff
@@ -6,6 +6,7 @@ def login_as(user)
       user.reload # because the user isn't re-queried via Warden
       super(user, scope: :user, run_callbacks: false)
     end
+
```
escowles: Super minor, but you could back out this whitespace change if you're making other changes.

tpendragon: I'd love to, but rubocop complains.

spec/factories/generic_files.rb
tpendragon: Done

spec/support/features/session_helpers.rb
escowles: OK, that's pretty weird that it passed before then.

PR #126, opened 2015-10-22
Adding MultiVolumeWork - fixes #107

### LINKED ISSUE #107: Create a multi-volume work curation concern

escowles:
* Adding MultiVolumeWork model
* Refactoring much of ScannedResource functionality to be shared
* Does not include UI for CRUD operations on linked ScannedResources


app/controllers/concerns/remote_record_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+          @presenter = show_presenter.new(curation_concern, current_ability)
+        end
+    end
+
+    def manifest_builder
+      ManifestBuilder.new(presenter)
+    end
+
+    def show_presenter
+      ScannedResourceShowPresenter
```
jpstroop: Should `ScannedResourceShowPresenter` be hard-coded in this module? You've defined `MultiVolumeWorkShowPresenter` as well, and `RemoteRecordController` is meant to be shared behavior for `MultiVolumeWorksController` and `ScannedResourcesController`, right? Maybe I'm missing something?

escowles: No -- that should be moved out to the controllers.  Right now they're basically the same, but that'll change as UI for MultiVolumeWork's members get fleshed out.

jpstroop: Do you want to move it into the respective controllers here? Or merge and deal when they're different? That was the only flag for me.

escowles: I'll go ahead and move it.  I also noticed that the updated `ScannedResourcesController#create` method that does error handling should be moved to the `RemoteMetadataController` module too, so I'll make both of those changes.

app/controllers/concerns/remote_record_controller.rb
```diff
@@ -0,0 +1,53 @@
+module RemoteRecordController
```
tpendragon: If this is just shared behavior between our various curation concern controllers, can we handle this with
```ruby
class CurationConcernController < ApplicationController
end
class ScannedBooksController < CurationConcernController
end
```

rather than multiple inheritance?

escowles: Yes, I think it'll work either way.  We could definitely have a CurationConcernController and have ScannedResourcesController and MultiVolumeWorkController subclass it.  Does that seem better than having the shared behavior in a module?

PR #129, opened 2015-10-22
Add authorized logic.

tpendragon:
Opening this PR largely for discussion about what the business logic around #authorized? should be. A custom role? Anyone who has campus_patron? Anyone who is a curation_concern_creator? Some other role?

### DIFF
--- a/app/models/concerns/pul_user_roles.rb
+++ b/app/models/concerns/pul_user_roles.rb
@@ -10,6 +10,6 @@ module PulUserRoles
   end
 
   def authorized?
-    roles.where(name: "authorized")
+    roles.where(name: "authorized").exists?
   end
 end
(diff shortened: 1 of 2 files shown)


tpendragon:
This logic might be done already via ability.rb

tpendragon:
As it stands, campus patrons can view curation concerns, creators can make them, and admins can manage roles. Should CAS logins just automatically get the patron group?

escowles:
I think having CAS logins automatically get the patron group makes sense.

Definitely another ticket, but having campus IP address ranges also get the group would be nice too.

jpstroop:
I'm working up a quick google doc around what I think our first pass at roles should be. Will plunk a link here in the next 5 minutes.

jpstroop:
https://goo.gl/znB0FR

PR #130, opened 2015-10-23
Adding UI for attaching ScannedResources to MultiVolumeWorks - closes #91

### LINKED ISSUE #91: Create Scanned Book on BookSet page

escowles:


app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -2,12 +2,23 @@ class CurationConcerns::CurationConcernsController < ApplicationController
...
+    begin
+      super
+    rescue StandardError => err
+      metadata_id = params[curation_concern_name]['source_metadata_identifier']
+      logger.debug "Error retrieving metadata for #{metadata_id}: #{err}"
+      flash[:alert] = "Error retrieving metadata for '#{metadata_id}'"
+      render :new, status: 500
+    end
+
+    return unless params[curation_concern_name][:parent_id]
```
tpendragon: What happens if this gets triggered?

app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -2,12 +2,23 @@ class CurationConcerns::CurationConcernsController < ApplicationController
...
+    parent = ActiveFedora::Base.find(params[curation_concern_name][:parent_id], cast: true)
+    parent.members << curation_concern.reload
```
tpendragon: Should we wrap these interactions up in an "Actor", as CC does? I'm a bit ambivalent, as I like seeing the logic in the controller action, but there may be strengths to following a similar pattern. Thoughts?

app/views/curation_concerns/base/_form_required_information.html.erb
```diff
@@ -1,3 +1,7 @@
+<% unless params[:parent_id].blank? %>
+  <%= f.hidden_field :parent_id, value: params[:parent_id] %>
```
tpendragon: Is this necessary? I would assume the form would POST to the nested create action, which would have parent_id

app/presenters/multi_volume_work_show_presenter.rb
```diff
@@ -1,3 +1,11 @@
 class MultiVolumeWorkShowPresenter < CurationConcerns::WorkShowPresenter
   delegate :date_created, :viewing_hint, :viewing_direction, to: :solr_document
+
+  def scanned_resource_presenters
```
tpendragon: Maybe we should rename this to member_presenters

app/views/curation_concerns/multi_volume_works/_members.html.erb
```diff
@@ -0,0 +1,17 @@
+<% if @presenter.scanned_resource_presenters.present? %>
+  <div class="panel panel-default scanned_resources">
+    <div class="panel-heading">
+      <h2>Scanned Resources</h2>
+    </div>
+    <ul>
+      <% @presenter.scanned_resource_presenters.each do |res| %>
+        <li>
+          <%= link_to res.title, main_app.curation_concerns_scanned_resource_path(res.id) %>
```
tpendragon: Can we get a thumbnail too?

spec/models/multi_volume_work_spec.rb
```diff
@@ -59,10 +59,10 @@
 
   describe 'has scanned resource members' do
     before do
-      subject.scanned_resources = [scanned_resource1, scanned_resource2]
+      subject.members = [scanned_resource1, scanned_resource2]
```
tpendragon: #ordered_members ?

app/controllers/curation_concerns/curation_concerns_controller.rb
tpendragon: On this same line: #ordered_members?

tpendragon: What's the reasoning behind #reload?

app/controllers/curation_concerns/curation_concerns_controller.rb
tpendragon: Can this even get triggered? Seems like the route's missing

app/views/curation_concerns/base/_form_required_information.html.erb
escowles: Looks like it's using /concern/scanned_resources/ instead of /container/:parent/scanned_resources/ -- but I can fix that instead of just inserting the hidden parent_id field.

app/controllers/curation_concerns/curation_concerns_controller.rb
escowles: The #reload was necessary because the ScannedResources in memory weren't AF::Base objects for some reason.  Though that may have been related to an earlier setup adding them to #scanned_resources instead of #members, so I'll double check that it's still needed.

app/views/curation_concerns/multi_volume_works/_members.html.erb
escowles: We should probably be using `CurationConcerns::ParentContainer` and just have a ScannedResource partial for this.  I tried that but it wasn't working because the partial expected a counter parameter.  Can we punt that to a new ticket?

tpendragon:
Having a tough time getting this to run locally, not sure if it's my setup or what

escowles:
I just walked through setting up a clean instance, following the README instructions, and the tests passed for me.

Here's what I'm doing after the README instructions to test manually:
* `rails s`
* Go to http://localhost:3000/users/auth/cas and login with CAS
* `rails c`
* `irb> User.last.roles << Role.first_or_create(name: "admin")`
* Go to http://localhost:3000/concern/multi_volume_works/new
* Fill in Title/Access/Use and click "Create Multi volume work" button
* Click "Attach a Scanned Resource" button
* Fill in Title/Access/Use and click "Create Scanned Resource"
* Find the MultiVolumeWork, and the ScannedResource shows up in the Scanned Resources section

Does that not work for you?  Are you getting any error messages or anything?

tpendragon:
Got it to work

PR #134, opened 2015-10-23
Add manifest ability specs.

tpendragon:


tpendragon:
@escowles Can you review? It's complaining about coverage, but only because total lines went down.

spec/ability/curation_concern_creator_spec.rb
```diff
@@ -59,11 +62,14 @@
       it { should_not be_able_to(:create, ScannedResource.new) }
       it { should_not be_able_to(:create, FileSet.new) }
       it { should_not be_able_to(:read, campus_only_scanned_resource) }
+      it { should_not be_able_to(:manifest, campus_only_scanned_resource) }
       it { should_not be_able_to(:pdf, campus_only_scanned_resource) }
       it { should_not be_able_to(:update, scanned_resource) }
       it { should_not be_able_to(:destroy, scanned_resource) }
+      it { should_not be_able_to(:manifest, campus_only_scanned_resource) }
```
escowles: Should be just `scanned_resource` right?

escowles:
Looks good other than that one small thing.  Also, add #139 to just add coverage for those two lines.

tpendragon:
Working on doing this without persisting those resources..

PR #138, opened 2015-10-23
Add vary header

### LINKED ISSUE #137: Pressing back displays JSON

tpendragon:
This fixes pushing back returning JSON. Curation Concerns doesn't have
this problem, not sure why we have it.

Closes #137


escowles:
Weird, but fixes the problem, so :+1:

PR #143, opened 2015-10-26
Return parents when searching for child metadata.

### LINKED ISSUE #93: Labels for a bookset's scanned books can be searched and will return the bookset

tpendragon:
Closes #93

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -1,6 +1,6 @@
 class CatalogController < ApplicationController
   include CurationConcerns::CatalogController
-  self.search_params_logic += [:hide_parented_resources]
+  self.search_params_logic += [:hide_parented_resources, :join_from_parent]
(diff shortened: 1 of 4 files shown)


tpendragon:
This solr query might not be very performant, but I couldn't get anything else to work

escowles:
I guess the question is whether we'd rather have bad search-time performance (doing the join) or bad index-time performance (loading the children and indexing their values in the parent record).

tpendragon:
I wish I could get it to work with a filter query

tpendragon:
@escowles Yeah. I suspect since this is an ingest tool that we're less worried about query performance we can't even measure yet. @jpstroop?

tpendragon:
This breaks the show view for scanned resources with parents. I'll work on that after lunch.

tpendragon:
Fixed, ready for review

PR #147, opened 2015-10-27
Adding state property - fixes #22

### LINKED ISSUE #22: There should be an interface that displays all things that aren't complete

escowles:


app/controllers/catalog_controller.rb
```diff
@@ -234,4 +238,8 @@ class CatalogController < ApplicationController
     # mean") suggestion is offered.
     config.spell_max = 5
   end
+  def admin_state_facet
+    return unless current_user && (current_user.curation_concern_creator? || current_user.admin?)
```
tpendragon: Can we move this to ability.rb and use cancan?

app/models/concerns/common_metadata.rb
```diff
@@ -15,6 +15,11 @@ module CommonMetadata
     property :source_metadata,
              predicate: ::RDF::URI.new('http://library.princeton.edu/terms/source_metadata'),
              multiple: false
+    property :state,
+             predicate: ::RDF::URI.new('http://fedora.info/definitions/1/0/access/objState'),
+             multiple: false do |index|
```
tpendragon: Indenting is a little odd here, surprised rubocop ignored it.

app/presenters/curation_concerns_show_presenter.rb
```diff
@@ -0,0 +1,7 @@
+class CurationConcernsShowPresenter < CurationConcerns::WorkShowPresenter
+  delegate :date_created, :viewing_hint, :viewing_direction, :state, to: :solr_document
+
+  def state_badge
+    StateBadge.new(solr_document).render
```
tpendragon: Gorgeous

app/models/concerns/common_metadata.rb
escowles: Yeah, the previous property definitions are indented so their arguments are in a block (rubocop seems to recognize and allow this), but then the block must be indented two spaces.

Related, maybe we should define some RDF vocabs for these predicates, and then the property definitions might be shorter enough to fit on one line?

tpendragon: > maybe we should define some RDF vocabs for these predicates

:+1: Is the fedora one not defined somewhere up the stack already?

app/controllers/catalog_controller.rb
tpendragon: Thinking something like

```ruby
if can? :manage_state, SolrDocument # or some other model...
  ...
end
```

tpendragon: That should let you remove the Devise/current_user stuff added in this PR too.

app/validators/state_validator.rb
```diff
@@ -0,0 +1,12 @@
+class StateValidator < ActiveModel::Validator
+  def validate(record)
```
tpendragon: Can we use an inclusivity validator for this? The error messages tend to be pretty. Something similar to how ViewingHintValidator works.

app/views/curation_concerns/base/_form_state.html.erb
```diff
@@ -0,0 +1,13 @@
+<fieldset id="set-state">
+  <legend>State</legend>
+  <div class="form-group">
+    <label class="radio">
```
tpendragon: Can we use simple form builders for these? Might push off some of the view branches

app/controllers/catalog_controller.rb
escowles: Strangely, `can? :edit, ScannedResource` returns true for admin, curation_concern_creator and anonymous users.

tpendragon: Yeah, you need an instance of ScannedResource to check against for edit.

escowles: Hmm, different problem with `can? :edit, ScannedResource.new`: works for admin/curation_concern_creator, but it results in a 404 for anonymous users.

tpendragon: Maybe can? :create, ScannedResource ?

tpendragon: We may want an explicit permission for this - who can review may be different than who can create

escowles: `can? :create, ScannedResource` works, so I think we should go with that.  We can create a new permission for review later on if creators and reviewers really are separate.

app/models/concerns/common_metadata.rb
escowles: The Fedora 4 ontology is in RDF.rb, but not the Fedora 3 ontologies.  I'll add stub vocabs for it and the PUL/OpaqueNamespace terms.

PR #148, opened 2015-10-29
Sending email and minting ARKs when records are completed

### LINKED ISSUE #19: There should be a way to mark an item as complete, at which point it will generate an Ark and email CAMS

escowles:
Fixes #19


config/config.yml
```diff
@@ -1,4 +1,7 @@
 defaults: &defaults
+  email:
+    from_address: plum@princeton.edu
+    reviewer_address: cams@princeton.edu
```
escowles: Wild guesses at the right email addressed to use here.

spec/models/scanned_resource_spec.rb
```diff
@@ -174,4 +174,33 @@
       expect(subject.valid?).to eq false
     end
   end
+
+  describe "#check_completion" do
+    it "completes record when state changes to 'complete'" do
+      allow(subject).to receive("state_changed?").and_return true
+      subject.state = 'complete'
+      expect(subject).to receive(:complete_record)
+      expect { subject.check_completion }.to change { ActionMailer::Base.deliveries.count }.by(1)
```
escowles: This test fails for me -- but the corresponding action works when I use the app locally (the email is dumped to the Rails log instead of sent).

spec/models/scanned_resource_spec.rb
```diff
@@ -174,4 +174,33 @@
...
+      expect(subject).not_to receive(:complete_record)
+      expect { subject.check_completion }.not_to change { ActionMailer::Base.deliveries.count }
+    end
+    it "does not overwrite existing identifier" do
+      allow(subject).to receive("state_changed?").and_return true
+      subject.state = 'complete'
+      subject.identifier = '1234'
+      expect(subject).to receive(:complete_record)
+      expect(subject).not_to receive("identifier=")
+      expect { subject.check_completion }.to change { ActionMailer::Base.deliveries.count }.by(1)
```
escowles: Same as above: this test fails, but the functionality appears to be working for me when I run the app.

config/config.yml
jpstroop: reviewer might be a list?

escowles: There are a few options here: a server-managed list, or a list that could be in the to/cc of a single email, or separate emails sent to each member.  A server-managed list would decouple updating the list from updating the app.

config/initializers/ezid.rb
```diff
@@ -0,0 +1,7 @@
+unless Rails.env.production?
```
tpendragon: Maybe we should remove this and just make sure the EZID stuff is configured in environment variables for staging/production?

spec/models/scanned_resource_spec.rb
tpendragon: Left a note in Slack, but for completion: The problem is the mock on :complete_record above. It replaces that method so it doesn't call anything. Change to something like:

```ruby
  describe "#check_completion", vcr: { cassette_name: "ezid" } do
    it "completes record when state changes to 'complete'" do
      allow(subject).to receive("state_changed?").and_return true
      allow(subject).to receive(:complete_record).and_call_original
      subject.state = 'complete'
      expect { subject.check_completion }.to change { ActionMailer::Base.deliveries.count }.by(1)
      expect(subject).to have_received(:complete_record)
    end
```

tpendragon: That being said, this is an integration test already - is the fact that #complete_record is called important?

config/initializers/ezid.rb
escowles: We'd also need to setup test (assume we can just add them to `circle.yml`).

tpendragon: Oh, I just meant the "unless Rails.env.production?" piece.

escowles:
I've updated the PR to address @tpendragon's comments, and the tests now pass because I'm not stubbing out the method I'm trying to test.  But how to support multiple reviewers getting email is unresolved.

config/config.yml
tpendragon: Can we just send it to everyone who can review?

jpstroop:
I wonder if reviewer should be a role (and then ask reviewers get an email)? I propose we merge  this as-is and bring it up as a question  when you guys are here.

jpstroop:
I wonder if reviewer should be a role (and then all reviewers get an email)? I propose we merge  this as-is and bring it up as a question  when you guys are here.

escowles:
:+1: -- I've created #153 as a placeholder question for figuring that out.

jpstroop:
One thing: are we sending any metadata to EZID? Does the gem let us do that? They complained to me a year or two ago that none of our ARKs have even a title associated with them...I said we could probably make it happen as part of our Hydra migration.

I don't think we need to go crazy, but http://ezid.cdlib.org/doc/apidoc.html#profile-dc, but creator, title, and publisher would be nice.

escowles:
This ticket doesn't set any metadata -- but the ezid-client gem does allow us to do that.  So we should be able to include basic DC fields.

jpstroop:
OK. I'll file a separate issue. I'd rather @tpendragon merge this.

jpstroop:
#154

jpstroop:
#154

tpendragon:
@jpstroop Are you okay with just emailing to that one address?

tpendragon:
Ah, just saw #153

Issue #62, opened 2015-09-10
Books and pages should have form fields for IIIF properties. 


jpstroop:
Follow on from #50.

jpstroop:
Not done from #50:

__Scanned Resources need:__

Viewing Direction (`sc:viewingDirection`). The value/object must be one of:
 * left-to-right (`sc:leftToRightDirection`) __(default)__
 * right-to-left (`sc:rightToLeftDirection`)
 * top-to-bottom (`sc:toToBottomDirection`)
 * bottom-to-top (`sc:bottomToTopDirection`)

Viewing Hint (`sc:viewingHint`). The value/object must be one of:
 * individuals (`sc:individualsHint`) __(default)__
 * paged (`sc:pagedHint`)
 * continuous (`sc:continuousHint`)

PR #156, opened 2015-11-03
Sorting UI for Scanned Resources

tpendragon:
This allows scanned resources' FileSets to be re-ordered via JQuery-UI
Sortable.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem 'coffee-rails', '~> 4.1.0'
 
 # Use jquery as the JavaScript library
 gem 'jquery-rails'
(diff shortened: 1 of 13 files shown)


tpendragon:
Rebased.

tpendragon:
UI looks like:

screen shot 2015-11-03 at 11 27 14 am

jpstroop:
Do you want to add the ability to update properties about the page/book, similar to what's in that demo I mocked up (https://github.com/pulibrary/pages-editor-mockup), or merge and make separate issues?

jpstroop:
I guess that's #62, but it's probably something we should have for the week of the 30th.

PR #168, opened 2015-11-10
Add Universal Viewer

tpendragon:
The Universal Viewer is architected in such a way as to be incompatible
with the asset pipeline, so adding it to public is the best we can do
unfortunately.


escowles:
Would this make sense as a submodule instead of including directly in Plum?  Do we manually update it when necessary?

tpendragon:
I've had bad luck maintaining submodules in projects in the past, but I'm happy to give it a shot again.

tpendragon:
Maybe we need a bower configuration?

escowles:
I haven't had a lot of experience with submodules, so I'm not sure how much trouble they are.  Manually updating the code also seems like a hassle, so I'm open to whatever process works.

Updating the dependencies is another level to this, and I see some bower config files in here (public/universalviewer/.bower.json, public/universalviewer/.bowerrc).

kevinreiss:
FWIW @jpstroop and I had a terrible time with submodules on our original Hydra pilot project.

tpendragon:
@escowles Updating universalviewer should now mean `bower update` and then fix representative_media. Maybe I can clean up that second bit?

tpendragon:
@escowles With the last commit, "bower update" will update UniversalViewer, and everything else will fix itself. Might have to fix the specs, but that seems appropriate.

escowles:
:+1:  `bower update` and fixing specs sounds like a reasonable update process to me.

jpstroop:
:thumbs_up:

PR #169, opened 2015-11-11
Convert Coffeescript to ES6

tpendragon:
ES6 provides many of the desirable pieces of Coffeescript, and as it will be natively supported by browsers at some point it's probably a good idea to switch to it.

Until most browsers have support, sprockets-es6 will transpile to regular JS.


tpendragon:
Hold off, I think this breaks something.

tpendragon:
Okay, as soon as Circle passes this should be good to go. Added linting of ES6 to Circle too.

Issue #100, opened 2015-10-16
Text in form fields are hard to read


tpendragon:

tpendragon:
Closed as duplicate of #86

PR #171, opened 2015-11-12
Add Browse Everything for Server Uploads

tpendragon:
This enables a user to upload to a server share and then select the
files they want to be added to the work from there. Files uploaded this
way maintain order and uploads happen in the background.

### DIFF
--- a/app/decorators/io_decorator.rb
+++ b/app/decorators/io_decorator.rb
@@ -1,0 +1,9 @@
+class IoDecorator < Hydra::Derivatives::IoDecorator
+  def original_filename
+    original_name
+  end
+
+  def content_type
+    mime_type
+  end
+end
--- a/app/values/file_info.rb
+++ b/app/values/file_info.rb
@@ -1,0 +1,30 @@
+class FileInfo
+  attr_reader :file_info
+  delegate :[], :has_key?, to: :file_info
+  def initialize(file_info)
+    @file_info = file_info
+  end
+
+  def to_h
+    file_info
+  end
+
+  def file_name
+    file_info["file_name"]
+  end
+
+  def file_path
+    FilePath.new(file_info["url"])
+  end
+
+  def mime_type
+    file_info["mime_type"] || mime_type_from_extension
+  end
+
+  private
+
+    def mime_type_from_extension
+      extension = File.extname(file_path.clean).delete(".")
+      Mime::Type.lookup_by_extension(extension).to_s
+    end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ noid-minter-state
 /jetty
 /fits
 /coverage
+/staged_files
--- a/Gemfile
+++ b/Gemfile
@@ -92,3 +92,4 @@ gem 'sidekiq'
 gem "omniauth-cas"
 gem 'ezid-client'
 gem 'sprockets-es6'
+gem 'browse-everything', github: 'projecthydra-labs/browse-everything'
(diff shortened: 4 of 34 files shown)


jpstroop:
I'm :+1:, wouldn't mind @escowles having a look.

app/decorators/io_decorator.rb
```diff
@@ -0,0 +1,9 @@
+class IoDecorator < Hydra::Derivatives::IoDecorator
+  def original_filename
```
escowles: Does browse_everything want different methods for filename and mime type than what hydra-derivatives uses?  Seems strange that there would be a mismatch.

app/values/file_info.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def file_name
+    file_info["file_name"]
+  end
+
+  def file_path
+    FilePath.new(file_info["url"])
+  end
+
+  def mime_type
```
escowles: Should we allow `file_info['mime_type']` to override the extension-based mime type guessing?  I don't have any use case for doing that now, but it's always nice to be able to explicitly set the mime type when you know a more specific type than what the extension would get you.

spec/jobs/browse_everything_ingest_job_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe BrowseEverythingIngestJob, type: :job do
```
escowles: Looks like BrowseEverythingIngestJob is covered by the ScannedResourcesController test, so maybe we should remove this file if we don't have any unit tests for it?

escowles:
Two small things:
* I'm seeing a regression on #100 (form text is light gray again).  `bundle install` installed Blacklight 5.16.3, so maybe there is updated CSS causing this?
* Totally not important, but can we call it `browse_everything_files` or `staged_files` or something instead of `server_bucket`?

app/decorators/io_decorator.rb
tpendragon: No - it seems more like the difference between a rails uploaded file and a file on disk. Hydra::Works/CurationConcerns expects an uploaded file.

app/values/file_info.rb
tpendragon: As far as I can tell, `file_info` doesn't come with `"mime_type"`. I could add it with the idea that maybe some services do add it? It looks like maybe dropbox does?

tpendragon:
> 'm seeing a regression on #100 (form text is light gray again). bundle install installed Blacklight 5.16.3, so maybe there is updated CSS causing this?

Yeah, I saw this too. Changing our CSS to `body.form-control` fixes it, but I figured I'd do that in a different PR.

> Totally not important, but can we call it browse_everything_files or staged_files or something instead of server_bucket?

`staged_files` :+1:

app/values/file_info.rb
escowles: I just checked and it looks like dropbox [does provide it](https://github.com/projecthydra-labs/browse-everything/blob/master/spec/fixtures/vcr_cassettes/dropbox.yml), so it would be nice to support that.

app/decorators/io_decorator.rb
escowles: OK, that makes sense.

app/values/file_info.rb
tpendragon: @escowles Added

escowles:
Sorry, one more thing: can you should remove staged_files/color.tif ?  Not sure how that got in since staged_files is in .gitignore.

tpendragon:
Bah, added that entry to gitignore after renaming the directory. I'll take care of it when I get back.

tpendragon:
@escowles Done!

PR #176, opened 2015-11-13
Add Multi Volume Work Manifests

### LINKED ISSUE #146: IIIF Manifest for Multi-Volume Work should display scanned resources under it

tpendragon:
This adds a viewer and IIIF manifests for multi volume works.

Closes #146.


app/services/manifest_builder.rb
```diff
@@ -1,32 +1,61 @@
 class ManifestBuilder
-  attr_reader :record
+  attr_reader :record, :light
```
jpstroop: Is the idea behind `:light` that if true only references to the Manifests are serialized?

tpendragon: Yeah. It stops the canvases from the scanned resources being added to the document.

jpstroop: OK, just making sure I understood--that's awesome!

PR #181, opened 2015-11-24
Sending IIIF Auth service definitions in manifests

escowles:


app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -43,7 +54,15 @@ def presenter
     end
 
     def manifest_builder
-      ManifestBuilder.new(presenter, ssl: request.ssl?)
+      ManifestBuilder.new(presenter, ssl: request.ssl?, login_url: login_url, logout_url: logout_url)
```
tpendragon: Thoughts on just passing in the auth manifest here, rather than building it in the manifest builder?

tpendragon: Or even just services: [auth_service], so we leave it open for more services on the manifest.

escowles: That sounds good -- it would detangle ManfiestBuilder and AuthManifestBuilder, and let us plug in more services in the controller if that makes sense later.

PR #182, opened 2015-11-24
Implement Bulk Label Interface

### LINKED ISSUE #7: Easy interface to bulk add labels to a set of pages

### LINKED ISSUE #170: Jasmine for Javascript?
It's been a while since I've looked at the state of integrating Jasmine with Rails, but maybe it's in a good place now?

tpendragon:
This is an interface for bulk labeling/previewing/persisting those
labels. Closes #7, closes #170

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,8 @@ rules:
   semi:
     - 2
     - never
+  no-constant-condition:
+    - 0
 env:
   es6: true
   browser: true
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,7 @@ group :development, :test do
   gem 'rspec-rails'
   gem 'jettywrapper'
   gem "factory_girl_rails"
+  gem 'jasmine-rails'
 end
 
 group :development do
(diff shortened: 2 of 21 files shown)


tpendragon:
screen shot 2015-11-24 at 12 49 40 pm

This is what it looks like.

escowles:
This looks very nice.  I spun it up locally and played with it and everything worked as expected, except:
* the updates didn't trigger reindexing in Solr
* clicking Save didn't take me back to the ScannedResource (or we add a button to go back to the ScannedResource)

tpendragon:
> the updates didn't trigger reindexing in Solr

Are you sure? It's using the file set update action, this would be odd. Steps to reproduce?

escowles:
You're right -- it is updated Solr.  It's just that the ScannedResource view of the FileSets is using ebucore:filename instead of dc:title, so the display wasn't updated.  I've created #183 for using dc:title (if available).

tpendragon:
@escowles Added a link back to parent.

PR #184, opened 2015-11-25
Adding review state (closes #23)

### LINKED ISSUE #23: Be able to mark as needing review again

escowles:


app/views/curation_concerns/base/_form_state.html.erb
```diff
@@ -2,11 +2,16 @@
   <legend>State</legend>
   <div class="form-group">
```
tpendragon: Can simple form help here to reduce some of the logic?

escowles: I can generate the radio buttons pretty easily, e.g.:

```
  <% states = { pending: 'In progress and suppressed from public display',
                review: 'In need of review',
                complete: 'Finished work, visibility determined by Access Rights' } %>
  <%= f.input :state, collection: states.map {|k,v| [k.to_s.titleize + ": " + v,k]}, as: :radio_buttons %>
```

But how do we generate the badges?

jpstroop: This seems pretty clean to me; what logic is of concern?

tpendragon: Just concerning that we have to update the checked value postfix conditional based on state for each state there is, which I thought simple form might be able to handle.

tpendragon: > But how do we generate the badges?

That's a good point. I'd have to fiddle with it, might not be able to.

tpendragon: Maybe as simple as a pending? method?

PR #196, opened 2015-12-01
Using background mail delivery, and showing identifier in views

escowles:


app/views/curation_concerns/base/_form_state.html.erb
```diff
@@ -1,20 +1,22 @@
...
-      In need of review
-    </label>
-    <label class="radio">
-      <%= f.radio_button :state, 'complete', checked: curation_concern.state == 'complete' %>
-      <%= f.label 'Complete', class: 'label label-success', for: 'scanned_resource_state_complete' %>
-      Finished work, visibility determined by Access Rights
-    </label>
-  </div>
-</fieldset>
+<% unless curation_concern.new_record? %>
```
tpendragon: Probably should have some sort of spec to show that it won't show up for new records.

PR #219, opened 2015-12-03
Adding a state machine to manage states and transitions

### LINKED ISSUE #211: Review workflow states per 12/1 discussion
@escowles can you add the whiteboard photo?

See also #205

escowles:
Closes #211

### DIFF
--- a/app/models/concerns/common_metadata.rb
+++ b/app/models/concerns/common_metadata.rb
@@ -41,7 +41,7 @@ module CommonMetadata
     end
 
     def check_completion
-      complete_record if self.state_changed? && state == 'complete'
+      complete_record if state_changed? && state == 'complete'
     end
 
     private
--- a/app/models/state_workflow.rb
+++ b/app/models/state_workflow.rb
@@ -1,0 +1,56 @@
+# State-based workflow: The primary workflow is to start at pending, and
+# progress through metadata_review and final_review to complete.  There are
+# two exceptional workflows: between complete/takedown (for issues requiring
+# an item be suppressed), and between complete/flagged (for issues where an
+# item can remain accessible).
+class StateWorkflow
+  include AASM
+
+  def initialize(state)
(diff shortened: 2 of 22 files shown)


app/models/state_workflow.rb
```diff
@@ -0,0 +1,69 @@
...
+# two exceptional workflows: between complete/takedown (for issues requiring
+# an item be suppressed), and between complete/flagged (for issues where an
+# item can remain accessible).
+class StateWorkflow
+  include AASM
+
+  def initialize(state)
+    aasm.current_state = state.to_sym unless state.nil?
+
+    @state_descriptions = {
```
tpendragon: Move these to I18n?

app/presenters/state_badge.rb
```diff
@@ -1,41 +1,63 @@
 class StateBadge
+  include ActionView::Context
   include ActionView::Helpers::TagHelper
 
-  def initialize(solr_document)
-    @solr_document = solr_document
+  def initialize(type, state = nil)
+    @type = type.underscore
+    @state = StateWorkflow.new state
+    @state_classes = {
```
tpendragon: Extract to instance method

app/models/concerns/common_metadata.rb
```diff
@@ -41,7 +41,12 @@ def apply_remote_metadata
     end
 
     def check_completion
-      complete_record if self.state_changed? && state == 'complete'
+      return unless self.state_changed?
+      state_machine = StateWorkflow.new state_was
+      unless state_machine.aasm.states(permitted: true).include? state.to_sym
+        raise ArgumentError, "Cannot transition from #{state_was} to #{state}"
```
tpendragon: Instead of raising an error, move these checks to the state validator.

app/models/state_workflow.rb
escowles: Good idea -- I moved the label and desc to I18n.

PR #227, opened 2015-12-04
Update Curation Concerns

tpendragon:


spec/forms/curation_concerns/multi_volume_work_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe CurationConcerns::MultiVolumeWorkForm do
+  subject { described_class.new(work, nil) }
+  let(:work) { MultiVolumeWork.new }
+
+  it "doesn't initialize description" do
```
escowles: Did we decide that MultiVolumeWorks shouldn't have a `description`?

jpstroop:
They should.

spec/forms/curation_concerns/multi_volume_work_spec.rb
jpstroop: They should.

tpendragon: This shouldn't stop that. It just makes it not fill a new record with a default of an empty array.

PR #232, opened 2015-12-07
Combine bulk label and reorder views.

### LINKED ISSUE #217: Merge reorder view and bulk_label view
As discussed. Perhaps make the grey area above each box dragable, and anywhere else selectable.

tpendragon:
Some work on flash messages needs to be done, but this closes #217.


escowles:
The combined UI looks good, and I was able to renumber pages just fine.  But I haven't been able to get reordering working.  I get the error "Unable to save new order." as a flash message.

tpendragon:
Huh, I wonder if style fixes broke it. I'll take a look after lunch.

spec/views/curation_concerns/scanned_resources/reorder.html.erb_spec.rb
```diff
@@ -1,44 +0,0 @@
-require 'rails_helper'
```
escowles: It would be nice to keep these tests.

tpendragon: Good call. Moved these over.

tpendragon:
@escowles Could you try to debug what's going on with reordering on your side? I can't get it to break

PR #234, opened 2015-12-08
Add a link to parent from scanned resource

eliotjordan:
Adds a breadcrumb on a scanned resource that leads back to the parent multivolume work. Looks good with short work titles, but what happens when one or both have longer titles? 

screenshot 2015-12-08 11 08 04

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -36,3 +36,5 @@
 @import "jquery-ui/sortable";
 @import "jquery-ui/selectable";
 @import "browse_everything";
+
+@import "components/breadcrumb";
--- a/app/assets/stylesheets/components/breadcrumb.scss
+++ b/app/assets/stylesheets/components/breadcrumb.scss
@@ -1,0 +1,8 @@
+.breadcrumb {
+  background-color: white;
+  margin-bottom: 10px;
+  border-radius: 0px;
+  border-width: 1px;
+  border-style: solid;
+  border-color: rgb(238, 238, 238);
+}
--- a/app/assets/stylesheets/components/positioning.scss
+++ b/app/assets/stylesheets/components/positioning.scss
@@ -4,4 +4,5 @@ main {
 
 .main-header {
     margin: 0;
+    border-bottom: 0;
 }
--- a/app/helpers/multi_volume_work_helper.rb
+++ b/app/helpers/multi_volume_work_helper.rb
@@ -1,0 +1,13 @@
+module MultiVolumeWorkHelper
+  def multi_volume_work_page_header
+    h = content_tag(:h1, @presenter.title)
+    h += multi_volume_work_breadcrumb
+    h.html_safe
+  end
+
+  def multi_volume_work_breadcrumb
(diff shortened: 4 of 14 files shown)


eliotjordan:
A few alternatives. I flipped the color scheme as well.

screenshot 2015-12-08 11 28 19

eliotjordan:
screenshot 2015-12-08 11 26 49

tpendragon:
I think I like `parent`, definitely not ID. I'm also concerned about the length of the two titles added together.

app/presenters/scanned_resource_show_presenter.rb
```diff
@@ -7,4 +7,11 @@ def file_presenter_class
   def pending_uploads
     @pending_uploads ||= PendingUpload.where(curation_concern_id: id)
   end
+
+  def parent
+    ids = solr_document.fetch('ordered_by_ssim', [])
+    CurationConcerns::PresenterFactory.build_presenters(ids[0...1],
```
tpendragon: This is problematic for the "scanned resource in multiple works" case, yeah?

tpendragon:
I think this has to be done via a nested route to get the parent_id, because the scanned resource can be in multiple works

app/presenters/scanned_resource_show_presenter.rb
eliotjordan: That is definitely a problem. The link to the scanned resource from the multi-volume work could be something like this: ```/concern/container/:parent_id/scanned_resources/:id```

jpstroop:
Rather than "parent" what about truncating the title somehow (first _n_ characters + '...'?)

eliotjordan:
@jpstroop That could lead to some weird text in the link. Maybe that's OK. I prefer simple text like 'parent', but I'm open. Here's a random mvw title truncated at 15 characters looks like:

screenshot 2015-12-08 15 42 53

escowles:
In the bulk editor, we have a link called "Back to Parent".  I think it would be fine to do the same here (and in general I think it is best to keep these things consistent).

I can see the value of including at least some of the title, though having it on one line combined with the ScannedResource title is probably going to break down.  Maybe it would be better to put it on its own line with some text or formatting to set it off?  e.g., "Back to MultiVolumeWork".

jpstroop:
I don't think "parent" makes a ton of sense to users. Even "parent work"
would be a little better if we went that route.

eliotjordan:
How about something like this:

screenshot 2015-12-08 16 30 42

jpstroop:
My concern is that we're asking users (even if they are trained cataloging staff) to understand too much about the models. "ScannedResource",  "Parent [Work]", "Multi Volume Work" make sense to us developers as abstractions of books, maps, and multi-part sets thereof, but I'm not sure how much those model names should surface in the UI. It's different if you're working with simpler/less fussy models (e.g. in an IR context where a Thesis is a Thesis is a Thesis). This is why my instinct is to use titles/labels in the hierarchy; even truncated versions if necessary (and maybe the whole string is visible when you hover?).

eliotjordan:
Ready for another look.

Issue #201, opened 2015-12-01
Send mail from the local machine


jpstroop:
at least for now

escowles:
Closed by #240

PR #240, opened 2015-12-09
Sending state notifications based on role membership

### LINKED ISSUE #153: Who should be notified that records are ready for review?
* A mailing list?
* Several users personal email addresses?
* All users who have the `reviewer` role?
* Some other option?

escowles:
Closes #153 and #201

### DIFF
--- a/app/mailers/reviewer_mailer.rb
+++ b/app/mailers/reviewer_mailer.rb
@@ -1,6 +1,10 @@
 class ReviewerMailer < ApplicationMailer
-  def completion_email(curation_concern_id)
+  def notify(curation_concern_id, state)
     @curation_concern = ActiveFedora::Base.find(curation_concern_id)
-    mail(to: Plum.config[:email][:reviewer_address], subject: "[plum] #{@curation_concern.human_readable_type} #{@curation_concern.id} is complete")
+    @state = state
+
+    addresses = Role.where(name: "notify_#{state}").first_or_create.users.map(&:email)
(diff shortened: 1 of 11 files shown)


jpstroop:
I'm :+1: to this approach

app/mailers/reviewer_mailer.rb
```diff
@@ -1,6 +1,10 @@
 class ReviewerMailer < ApplicationMailer
-  def completion_email(curation_concern_id)
+  def notify(curation_concern_id, state)
     @curation_concern = ActiveFedora::Base.find(curation_concern_id)
-    mail(to: Plum.config[:email][:reviewer_address], subject: "[plum] #{@curation_concern.human_readable_type} #{@curation_concern.id} is complete")
+    @state = state
+
+    Role.where(name: "notify_#{state}").first_or_create.users.each do |u|
+      mail(to: "#{u.email}", subject: "[plum] #{@curation_concern.human_readable_type} #{curation_concern_id}: #{state}")
```
tpendragon: I don't know if it's a problem, but I'm moderately sure this'll send N emails. Would we rather do all of the to: in one line? Or as BCC?

jpstroop: Having been on these kinds of list before, it's definitely better if I can see the other people who received the message, so that I can start a reply-all thread if I need to.

escowles:
I've updated it to put all the addresses in the to: field of a single email.

PR #242, opened 2015-12-11
Structural Metadata Editor

### LINKED ISSUE #18: Structural metadata for books.

### LINKED ISSUE #225: Structural metadata should become IIIF ranges

tpendragon:
This feature works with the theory that structural metadata can be
abstracted into a hash that looks like the following:

```ruby
{
  "nodes": [
    {
      "label": "Chapter 1",
      "nodes": [
        {
          "proxy": "mcvaa9zja"
        }
      ]
    }
  ]
}
```

This hash is created by the frontend with `StructureParser`, is
converted into an ordered graph by `LogicalOrder`, and can be created
from an ordered graph by `LogicalOrderGraph`.

This hash is also used to create IIIF Manifest Ranges for the viewer.

Closes #225, closes #18.

### DIFF
--- a/app/controllers/curation_concerns/scanned_resources_controller.rb
+++ b/app/controllers/curation_concerns/scanned_resources_controller.rb
@@ -48,6 +48,17 @@ class CurationConcerns::ScannedResourcesController < CurationConcerns::CurationC
     @members = presenter.file_presenters
   end
 
+  def structure
+    @members = presenter.file_presenters
+    @logical_order = presenter.logical_order_object
+  end
+
+  def save_structure
+    curation_concern.logical_order.order = { "nodes": params["nodes"] }
+    curation_concern.save
+    head 200
+  end
+
   private
 
     def lock_manager
--- a/app/models/logical_order_base.rb
+++ b/app/models/logical_order_base.rb
@@ -1,0 +1,47 @@
+class LogicalOrderBase < ActiveFedora::Base
+  property :nodes, predicate: ::RDF::Vocab::DC.hasPart
+  property :head, predicate: ::RDF::Vocab::IANA['first'], multiple: false
+  property :tail, predicate: ::RDF::Vocab::IANA.last, multiple: false
+
+  def order=(order)
+    nodes_will_change!
+    head_will_change!
+    tail_will_change!
+    order = LogicalOrder.new((order), ::RDF::URI(rdf_subject))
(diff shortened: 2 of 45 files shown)


escowles:
I think this is good enough to start iterating on.  One easy change I think would be good is to make the "+" into a button with a label like "Add a Section" to make it more obvious.  But the editing UI works well already.

tpendragon:
Squashed commits, ready for review + merge.

app/controllers/curation_concerns/scanned_resources_controller.rb
```diff
@@ -48,6 +48,17 @@ def bulk_edit
...
+    @members = presenter.file_presenters
+    @logical_order = presenter.logical_order_object
+  end
+
+  def save_structure
+    curation_concern.logical_order.order = { "nodes": params["nodes"] }
+    curation_concern.save
+    head 200
+  end
+
```
escowles: I like the naming of the methods -- I wonder if we should rename bulk_edit/save_order to match?

app/models/logical_order_base.rb
```diff
@@ -0,0 +1,48 @@
...
+    # Delete old statements
+    subj = resource.subjects.to_a.select { |x| x.to_s.split("/").last.to_s.include?("#g") }
+    subj.each do |s|
+      resource.delete [s, nil, nil]
+    end
+    self.head = nil
+    self.tail = nil
+    resource << graph
+    # Set nodes so that hash URIs get persisted to Fedora.
+    self.nodes = nil
```
escowles: Is this needed?  Or does the next line cover it?

app/models/logical_order_base.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  # Not useful, slows down indexing.
+  def modified_date
+    nil
+  end
+
+  # Serializing head/tail/nodes slows things down CONSIDERABLY, and is not
+  # useful.
+  # @note This method is used by ActiveFedora::Base upstream for indexing,
```
escowles: This is very interesting -- I would not have expected the profile to contain the structure by default.

tpendragon: Yeah - had to do something similar for AF::Aggregations

jpstroop:
Sorry, late +1. I agree, this is a great start!

-Js

PR #245, opened 2015-12-14
Adding collections - closes #206

### LINKED ISSUE #206: Implement Collections
So that Objects can be added to one or more for synchronization w/ Collections pomegranate.

escowles:


spec/validators/exhibit_id_validator.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe ExhibitIdValidator do
+  subject { described_class.new }
+
+  describe "#validate" do
+    let(:errors) { double("Errors") }
+    before do
+      allow(subject).to receive(:existing_exhibit_ids).and_return(['slug1'])
```
tpendragon: Stubbing the object under test is usually bad joojoo. Can you go a level deeper?

tpendragon:
Can we add "Create a Collection" to the front page? Maybe as the first option?

tpendragon:
"Add to Collection" from the search results view pops up a modal which seems to immediately dissapear

spec/validators/exhibit_id_validator.rb
escowles: OK, I'll stub ActiveFedora::SolrService.query instead of the ids method.

tpendragon: If that turns out to be too complicated, maybe a query object to encapsulate it?

escowles:
I've added a button to create a Collection to the home page, stubbed the ActiveFedora::SolrService.query method instead of the ExhibitIdValidator, and suppressed links to the broken "Add to a Collection" modal.

tpendragon:
:+1: Will merge when it passes.

tpendragon:
@escowles Coverage. =( Looks like `collectible`

escowles:
Not sure why the feature test didn't cover that -- I've added a controller test to get it back to :100:

PR #246, opened 2015-12-14
Allow drag for generating structure.

tpendragon:


escowles:
:+1: -- the dragging works for me, and having an obvious drag anchor helps too.

tpendragon:
The only weird thing is creating, IE, chapter 2 means you have to drag two-levels deep to create the node, and then drag that created node back a level.

No solution for that though, maybe that's just how it is.

jpstroop:
(From Slack conversation:)
 * [ ] I wonder if bringing back the button you had near the top to create an arbitrary range might help mitigate your "chapter 2" problem?
 * [ ] probably better to add the new node at the top?
 * [ ] shift+click doesn't seem to work
 * [ ] should be able to collapse any structure I've created

tpendragon:
So I added the requested features. Right now the way the plugin works is as you drag an object down, it'll expand collapsed items so that you can sort it into the tree. Which makes sense, but maybe we should disable that feature? It makes the "button to add a section" more troublesome than helpful, I think.

I also wonder if maybe there should be a "section above me" button.

tpendragon:
Deploying this to staging to try out.

jpstroop:
I will try to look today after lunch

jpstroop:
Was too anxious :wink:.

I think we're barking up the right tree.

How about cleaning up the boxes:
 * [ ] add thumbnails
 * [ ] get rid of bullets (I assume I'm not the only one seeing them? See image below)
 * [ ] move the expand/collapse arrow to the right side (wouldn't want to accidentally click the 'x')
 * [ ] make the x seem more severe - maybe even a red bootstrap warning label? 
 * [ ] make the form box a little bigger

And add a notification when the save is successful / fails, and then call this good for Friday?

screen shot 2015-12-15 at 11 01 15 am

tpendragon:
@jpstroop "Form box"?

tpendragon:
@jpstroop Maybe the "x" should be on the right side, by itself? Seems like having to reach for it would be important.

jpstroop:
Sorry. The field where the user adds the label for the new range. Some of them get really long. E.g. search for Glinka in the PUDL.

jpstroop:
> Maybe the "x" should be on the right side, by itself? Seems like having to reach for it would be important.

We can try whatever, but I also found the expand/collapse button difficult to find and thought that might make more sense to the far right.

tpendragon:
@jpstroop Pushed up some styling changes, take a look at staging?

jpstroop:
:+1: I'm cool w/ the expand/collapse arrows now too. I think thumbnails will give us a good first pass for Friday and you should make this into a PR once they're added.

tpendragon:
I'm curious about how to do thumbnails - I'm concerned about size issues if I add a .panel-body. Any thoughts?

tpendragon:
@jpstroop Added thumbnails and pushed to staging - take a look?

tpendragon:
@jpstroop says :+1:, so squashed and ready for review.

PR #247, opened 2015-12-15
Adding viewing direction/hint - closes #62

### LINKED ISSUE #62: Books and pages should have form fields for IIIF properties.

escowles:
This gets the job done, but there is some weirdness with the label display that I wasn't able to improve with the various _html options:

screen shot 2015-12-15 at 11 32 34 am 2

I think this is probably related to simple_form doing:

``` html
<span>
  <input/>
  <label>text</label>
</span>
```

But Bootstrap wanting:

``` html
<label>
  <input/>
  text
</label>
```

### DIFF
--- a/app/assets/javascripts/bulk_label.es6
+++ b/app/assets/javascripts/bulk_label.es6
@@ -17,6 +17,7 @@ class BulkLabeler {
       $(this).data("current-value", $(this).val())
     })
     this.track_radio_buttons()
+    this.track_resource_form()
     this.setup_buttons()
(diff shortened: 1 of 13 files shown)


tpendragon:
We talked about this in standup today - I think this is a simple form configuration issue. Also, i had assumed these buttons would go in bulk edit with the rest of the IIIF settings, but maybe that's not the right place for it. @jpstroop?

jpstroop:
I was assuming bulk edit.

PR #258, opened 2015-12-17
Collection manifests.

### LINKED ISSUE #224: Implement collection manifests

tpendragon:
Closes #224

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -8,4 +8,26 @@ class CollectionsController < ApplicationController
   def presenter_class
     CollectionShowPresenter
   end
+
+  def manifest
+    respond_to do |f|
+      f.json do
+        render json: manifest_builder
+      end
+    end
+  end
+
+  private
+
+    def manifest_builder
+      ManifestBuilder.new(presenter, ssl: request.ssl?, services: AuthManifestBuilder.auth_services(login_url, logout_url))
+    end
+
+    def login_url
+      main_app.user_omniauth_authorize_url(:cas)
+    end
+
+    def logout_url
+      current_user.nil? ? nil : main_app.destroy_user_session_url
+    end
 end
--- a/app/presenters/collection_show_presenter.rb
+++ b/app/presenters/collection_show_presenter.rb
@@ -1,4 +1,34 @@
 class CollectionShowPresenter < CurationConcerns::CollectionPresenter
   include CurationConcerns::Serializers
   delegate :id, :title, :exhibit_id, to: :solr_document
+
+  def file_presenters
+    @file_sets ||= CurationConcerns::PresenterFactory.build_presenters(ordered_ids,
+                                                                       file_presenter_class,
+                                                                       current_ability)
+  end
+
+  def viewing_hint
+    nil
+  end
+
+  def logical_order
+    {}
+  end
+
+  def label
+    []
+  end
+
+  private
+
+    # TODO: Extract this to ActiveFedora::Aggregations::ListSource
+    def ordered_ids
+      Array(solr_document["member_ids_ssim"])
+    end
+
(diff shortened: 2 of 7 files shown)


app/controllers/collections_controller.rb
```diff
@@ -8,4 +8,26 @@ def form_class
...
+  end
+
+  private
+
+    def manifest_builder
+      ManifestBuilder.new(presenter, ssl: request.ssl?, services: AuthManifestBuilder.auth_services(login_url, logout_url))
+    end
+
+    def login_url
+      main_app.user_omniauth_authorize_url(:cas)
```
escowles: We are hard-coding the auth provider here and elsewhere.  It's beyond the scope of this ticket to fix that, but maybe we should have a new issue to start inspecting the omniauth providers instead?  That would be one thing we'd need to do to make it possible for someone else to use Plum with a different auth system.

app/presenters/collection_show_presenter.rb
```diff
@@ -1,4 +1,34 @@
...
+  private
+
+    # TODO: Extract this to ActiveFedora::Aggregations::ListSource
+    def ordered_ids
+      Array(solr_document["member_ids_ssim"])
+    end
+
+    # Override this method if you want to use an alternate presenter class for the files
+    def file_presenter_class
+      ScannedResourceShowPresenter
```
escowles: There's a disconnect here with #**file**_presenter_class returning a **ScannedResource**ShowPresenter.  Should we rename the method?

config/routes.rb
```diff
@@ -38,6 +38,13 @@
     end
   end
 
+  resources :collections do
+    member do
+      get :manifest, defaults: { format: :json }, as: "manifest_curation_concerns"
+      get :manifest, defaults: { format: :json }
```
escowles: Duplicate route definition

app/presenters/collection_show_presenter.rb
tpendragon: It should probably be #member_presenter_class/#member_presenters, but that means we should probably fix it in CC.

config/routes.rb
tpendragon: It's actually on purpose, because the helpers resolve to two things in different places, since I could never get polymorphic_url to work in the builder. I could fix it by passing in the manifest url, maybe?

app/presenters/collection_show_presenter.rb
escowles: I created a CC issue for this: https://github.com/projecthydra-labs/curation_concerns/issues/564

config/routes.rb
escowles: Can you pass multiple things to `as`?  I think it's fine to leave it as-is if both helpers are needed in different places, but it does look strange.

tpendragon: Nah, you can't, but I should probably find a solution to there being two helpers.

app/controllers/collections_controller.rb
tpendragon: I think the whole builder infrastructure needs redesigned to allow for more flexibility to handle that kind of stuff. The churn in this file is kind of unreal.

app/presenters/collection_show_presenter.rb
tpendragon: Do we want to delay merging this for that?

config/routes.rb
tpendragon: Found a solution, running specs and then pushing it up shortly.

app/presenters/collection_show_presenter.rb
escowles: No, I think it's fine to let that play out however and update when that's resolved.

escowles:
I went ahead and added an issue for this: https://github.com/pulibrary/plum/issues/263

Not a high priority, but a good thing to do at some point.

PR #269, opened 2015-12-17
Refactor manifest builder

tpendragon:
This bases the manifest builder on a series of "sub-builders", which actually mutate the manifest it eventually generates. It accepts extra builders through the services parameter, and could be refactored later to accept something like a BuilderFinder, if we find it useful. The common API of Builder#apply(manifest) should lower the churn on this file by a fair bit.

### DIFF
--- a/app/services/auth_manifest_builder.rb
+++ b/app/services/auth_manifest_builder.rb
@@ -2,21 +2,37 @@
 # @see http://iiif.io/api/auth/0.9/
 class AuthManifestBuilder
   def self.auth_services(login_url, logout_url)
-    {
+    new(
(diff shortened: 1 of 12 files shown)


tpendragon:
There's a commit in here to fix manifests for structure too, which we should get in before tomorrow's demo.

jpstroop:
Cool 😎, nice and tidy. Remind me to _never_ show you the code that generates manifests for LAE

jpstroop:
Sounds like you want to keep those commits separate?

tpendragon:
> Sounds like you want to keep those commits separate?

Yup

PR #270, opened 2015-12-17
Apply viewing hint to sequence as well as manifest

tpendragon:


tpendragon:
@escowles @jpstroop Since we're all gone for the night, can someone merge this in the morning? Paged viewing in UV doesn't work without it. =(

PR #272, opened 2015-12-18
Fixed position flash alerts

### LINKED ISSUE #256: Fixed position flash div
Add a fixed position div for flash messages to the bulk edit view. Currently, user has to scroll up to the top of the page when there are lots of files in a scanned resource.

Should this be expanded to all pages?

eliotjordan:
Positions flash alert div in a fixed position at the top of the page. Also moves javascript alerts to the same position. Closes #256.

### DIFF
--- a/app/assets/javascripts/sortable.es6
+++ b/app/assets/javascripts/sortable.es6
@@ -40,7 +40,6 @@
         } else {
           this.flash.set("danger", "Unable to save new order.")
         }
-        this.disable()
       }
     }
 
--- a/app/assets/javascripts/flash.es6
+++ b/app/assets/javascripts/flash.es6
@@ -1,7 +1,7 @@
 /* exported Flash */
 class Flash {
   constructor() {
-    this.element = $("*[data-reorder-action='flash']")
+    this.element = $(".alert-js")
     this.element.children(".close").click(
       () => {
         this.element.hide()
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/sortable.es6
```diff
@@ -40,7 +40,6 @@
         } else {
           this.flash.set("danger", "Unable to save new order.")
         }
-        this.disable()
```
eliotjordan: Not sure why this doesn't work, but it raises an error in the browser. Removing it doesn't seem to change functionality.

jpstroop:
I think we should wait on this now until after the demo :ok_hand: ?

tpendragon:
:+1: to waiting. `sortable.es6` is dangerous territory.

tpendragon:
@jpstroop @escowles This changes -all- flash alerts to be fixed at the top. Is that desirable behavior?

@eliotjordan At the very least it seems like the height's a little big.

escowles:
:+1: to making all flash alerts fixed at the top -- doing this consistently across all views seems better than having some be fixed and some inline.

eliotjordan:
@tpendragon Height suggestion?

tpendragon:
@eliotjordan Maybe a padding-top/padding-bottom on the .alert of only 5px, instead of 25?

eliotjordan:
@tpendragon I'll give it a go.

eliotjordan:
Padding reduced to 5px.

PR #283, opened 2015-12-22
Adding flag method

### LINKED ISSUE #266: Create actions for state transitions
Reviewing the roles and actions in #213, there are a few state transitions that users should be allowed to perform (e.g., flagging for review), even though they shouldn't have permission to edit the record.

To isolate these state transitions and allow users to perform them, we should create actions for at least:
* flag (everyone except anonymous)
* final_review (editors/image editors)

escowles:
Closes #266

### DIFF
--- a/app/controllers/curation_concerns/curation_concerns_controller.rb
+++ b/app/controllers/curation_concerns/curation_concerns_controller.rb
@@ -33,6 +33,23 @@ class CurationConcerns::CurationConcernsController < ApplicationController
     end
   end
 
+  def flag
+    curation_concern.state = 'flagged'
+    note = params[curation_concern_name][:workflow_note]
+    curation_concern.workflow_note = curation_concern.workflow_note + [note] unless note.blank?
+    if curation_concern.save
+      respond_to do |format|
+        format.html { redirect_to [main_app, curation_concern], notice: "Resource updated" }
+        format.json { render json: { state: state } }
+      end
+    else
+      respond_to do |format|
+        format.html { redirect_to [main_app, curation_concern], alert: "Unable to update resource" }
+        format.json { render json: { error: "Unable to update resource" } }
+      end
+    end
+  end
+
   private
 
     def presenter
--- a/app/presenters/curation_concerns_show_presenter.rb
+++ b/app/presenters/curation_concerns_show_presenter.rb
@@ -1,8 +1,9 @@
 class CurationConcernsShowPresenter < CurationConcerns::WorkShowPresenter
   delegate :date_created, :viewing_hint, :viewing_direction, :state, :type, :identifier, :workflow_note, :logical_order, :logical_order_object, to: :solr_document
+  delegate :flaggable?, to: :state_badge_instance
 
   def state_badge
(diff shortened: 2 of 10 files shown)


escowles:
For some reason, appending notes to workflow_note works in the [model test](https://github.com/pulibrary/plum/pull/283/files#diff-1b2802872554842ed497fd0c7d813f76R27), but fails in the [controller test](https://github.com/pulibrary/plum/pull/283/files#diff-6259036f87e12d3642d5bcbcc3a36b15R360)

app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -33,6 +33,23 @@ def deny_access(exception)
     end
   end
 
+  def flag
+    curation_concern.state = 'flagged'
+    note = params[curation_concern_name][:workflow_note]
+    curation_concern.workflow_note = curation_concern.workflow_note + [note] unless note.blank?
```
escowles: This is ungainly, but works, unlike the previous attempt:

```
curation_concern.workflow_note << note unless note.blank?
```

tpendragon: Oh. Yeah. https://github.com/ActiveTriples/ActiveTriples/issues/19 =( I should have looked at the code, this has been an issue for a long time.

escowles: OK, glad to know it's not just me

app/presenters/curation_concerns_show_presenter.rb
```diff
@@ -5,6 +5,10 @@ def state_badge
     StateBadge.new(type, state).render
   end
 
+  def flaggable?
+    StateBadge.new(type, state).flaggable?
```
tpendragon: Extract `StateBadge.new(type, state)` to a private method, and delegate flaggable? Also used in the above method.

app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -33,6 +33,23 @@ def deny_access(exception)
     end
   end
 
+  def flag
```
tpendragon: So technically I can POST a flag even if it's not flaggable. It's not exactly a game-changer, but can we check for it here?

escowles: If the resource isn't flaggable, then the save will fail because of the state validation -- there's a test for this to make sure it won't let you flag a pending resource: https://github.com/pulibrary/plum/pull/283/files#diff-6259036f87e12d3642d5bcbcc3a36b15R386

tpendragon: :+1:

tpendragon: State machines are the best.

app/presenters/curation_concerns_show_presenter.rb
escowles: :+1:  -- I'll update the PR shortly.

PR #288, opened 2016-01-04
Fix typo in message.

demiankatz:
No big deal, obviously, but I noticed it, so I fixed it. :-)


tpendragon:
That one always gets me...

Thanks for the PR!

PR #298, opened 2016-01-05
When creating PDF downscale when possible.

tpendragon:
We may have some tweaking to do here to handle the size of the created
PDFs and how long it takes to generate them.


app/values/scanned_resource_pdf/canvas_downloader.rb
```diff
@@ -17,22 +18,22 @@ def layout
...
+    private
 
       def canvas_url
-        "#{canvas.url}/full/!#{max_width},#{max_height}/0/#{quality}.jpg"
+        "#{canvas.url}/full/#{max_width},#{max_height}/0/#{quality}.jpg"
       end
 
       def max_width
-        792
+        [(612 * 1.5).round, canvas.width].min
```
escowles: Why the math here?  Why not just use 918 and 1188?

tpendragon: I liked having the dimensions for letter in there. I'd pull it right out of the PDF document size if I could, but I couldn't find it.

jpstroop: Unless you wanted to make the scale factor configurable. Regardless, would it make sense to make these numbers constants so that we know how they were determined?

tpendragon: :+1: At least to make it more clear.

app/values/scanned_resource_pdf/canvas.rb
```diff
@@ -1,5 +1,7 @@
 class ScannedResourcePDF
   class Canvas
+    LETTER_WIDTH = 612
```
escowles: OK, now I see where 612 and 792 are coming from -- number of points in a 8.5x11 page.  Perhaps I should have expected the units to be points (because PostScript), but I was expecting pixels instead.  Maybe a comment here or at canvas_downloader#scale_factor making the units explicit would be a good idea?

tpendragon: You know, I don't really know what the scale factor is doing to those images if it's in points. I was assuming pixels too, but you're right.

PR #301, opened 2016-01-05
Add a cache-ready resource identifier.

tpendragon:
This is a stepping stone towards caching PDFs.

### DIFF
--- a/app/values/resource_identifier/solr_repository.rb
+++ b/app/values/resource_identifier/solr_repository.rb
@@ -1,0 +1,66 @@
+class ResourceIdentifier
+  ##
+  # Repository for querying solr for a single or multiple IDs and getting back a
+  # document object which can return its dependants and timestamps.
+  class SolrRepository
+    class << self
+      # @param [Array<Integer> OR Integer] id ID or array of IDs of records to
+      #   query for.
+      # @return [Record, #timestamp, #dependant_ids] Solr-record powered record
+      #   which can return its timestamp and dependant ids.
+      def find(id)
+        id = Array.wrap(id)
+        return CompositeRecord.new({}) if id.blank?
+        records = solr.query(solr_query(id), fl: "timestamp member_ids_ssim", sort: "timestamp desc").map do |r|
+          Record.new(r)
+        end
+        CompositeRecord.new(records)
+      end
+
+      private
+
+      def solr_query(id)
+        "id:(#{id.join(' OR ')})"
+      end
+
+      def solr
+        ActiveFedora::SolrService
+      end
+    end
+
+    class Record
+      attr_reader :record
+      delegate :fetch, :[], to: :record
+      def initialize(record)
+        @record = record
+      end
+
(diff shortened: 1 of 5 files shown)


app/values/resource_identifier/solr_repository.rb
```diff
@@ -0,0 +1,62 @@
...
+  class SolrRepository
+    class << self
+      # @param [Array<Integer> OR Integer] id ID or array of IDs of records to
+      #   query for.
+      # @return [Record, #timestamp, #dependant_ids] Solr-record powered record
+      #   which can return its timestamp and dependant ids.
+      def find(id)
+        id = Array.wrap(id)
+        return CompositeRecord.new({}) if id.blank?
+        records = solr.query("id:(#{id.join(' OR ')})").map do |r|
```
escowles: This seems a little heavy to me -- if there are a large number of FileSets, then loading them all, even from Solr, seems like a lot of data to move around to find the most recent timestamp.  Can't we do a Solr query like "id:X OR generic_work_ids_ssim:X", sort by timestamp, and get the timestamp from the first record?

tpendragon: The one time this might be an issue is deletes, but we can give it a shot and see if that becomes an issue.

tpendragon: (By deletes I mean it orders A, B, C, and C is most recently updated, then B is deleted - key won't change, but the state should.)

escowles: Wouldn't the A get reindexed if one of the members is removed?  I'm going to test that out, because it's something we need to account for one way or the other.

tpendragon: I'm adding the unit test now to see. A wouldn't, but the list_source would.

escowles: Manual testing shows A getting reindexed.

tpendragon: That's...sort of concerning.

app/values/resource_identifier/solr_repository.rb
```diff
@@ -0,0 +1,66 @@
...
+  class SolrRepository
+    class << self
+      # @param [Array<Integer> OR Integer] id ID or array of IDs of records to
+      #   query for.
+      # @return [Record, #timestamp, #dependant_ids] Solr-record powered record
+      #   which can return its timestamp and dependant ids.
+      def find(id)
+        id = Array.wrap(id)
+        return CompositeRecord.new({}) if id.blank?
+        records = solr.query(solr_query(id), fl: "timestamp member_ids_ssim", sort: "timestamp desc").map do |r|
```
tpendragon: To make this more efficient we would have to make filesets reindex their parents when they're deleted.

escowles: That seems like a good thing to do anyway, since we would also want to update the members list.

PR #318, opened 2016-01-07
Pulling in FileSetControllerBehavior#update method to customize destination after submit

### LINKED ISSUE #282: After editing a file from the bulk edit page, the user should be directed back to the bulk edit page
At the moment, the user redirected to the un-nested scanned resource page.

escowles:
Closes #282

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -27,7 +27,7 @@ GIT
 
 GIT
   remote: git://github.com/projecthydra-labs/curation_concerns.git
-  revision: 08008bf9073ff0fe4b7d60d9327ab4b643825fe8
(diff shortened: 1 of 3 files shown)


tpendragon:
This is a lot to take responsibility for. Can pull the destination out into a method in CC and then override that?

escowles:
Yeah, I think the pattern in the base CurationConcernsController  is much better -- theres an `after_create_response` method that just handles the redirect/json output.  So I can refactor the update method in CC to have an analogous method and then only override it.

escowles:
Updated to use the new `FileSetsController#after_update_response` method.

PR #326, opened 2016-01-08
Adding collection and language facets - closes #208

### LINKED ISSUE #208: Additional facets
Facet for collection, language...what else?

escowles:


spec/models/scanned_resource_spec.rb
```diff
@@ -315,6 +315,7 @@
     end
     it "marshals logical order into solr" do
       subject.logical_order.order = params
+      subject.save
```
tpendragon: Can't solrize without it being persisted?

escowles: This works, but I admit it's pretty weird.  It also explains why it's fine in the app, because of course it saves before solrizing.

PR #328, opened 2016-01-08
Fix thumbnail sizes in bulk labeler

tpendragon:


jpstroop:
It looks like the images are getting upsampled

tpendragon:
Are you seeing that with this PR?

tpendragon:
I ask because images shouldn't be able to escape the box with this PR.

PR #335, opened 2016-01-11
Indexing collection slug - closes #332

### LINKED ISSUE #332: Allow for search by collection slug

escowles:


spec/controllers/catalog_controller_spec.rb
```diff
@@ -67,11 +67,23 @@
...
+      expect(document_ids).to eq [resource.id]
+    end
+  end
 
-    def document_ids
-      assigns[:document_list].map do |x|
-        x["id"]
-      end
+  def document_ids
+    puts assigns[:docuement_list]
```
tpendragon: Delete?

app/models/scanned_resource.rb
```diff
@@ -14,7 +14,10 @@ def to_solr(solr_doc = {})
       doc[ActiveFedora::SolrQueryBuilder.solr_name("ordered_by", :symbol)] += send(:ordered_by_ids)
       doc[ActiveFedora::SolrQueryBuilder.solr_name("logical_order", :symbol)] = [logical_order.order.to_json]
       doc[ActiveFedora::SolrQueryBuilder.solr_name("logical_order_headings", :stored_searchable)] = logical_order.object.each_section.map(&:label)
-      doc[ActiveFedora::SolrQueryBuilder.solr_name("collection", :facetable)] = in_collections.map(&:title)
+      in_collections.each do |col|
```
tpendragon: We'll have to be careful with this - if things start to slow down a lot, this may be a culprit.

escowles: Yes, if this gets too slow, we may want to think about alternative ways to handle this.  E.g., we could probably use the presenter to pull the collection info from Solr.

PR #356, opened 2016-01-18
Enable MVW Reordering.

### LINKED ISSUE #355: I should be able to reorder multi volume works

tpendragon:
Closes #355


app/controllers/curation_concerns/curation_concerns_controller.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
```
escowles: Want to make a ticket to refactor this controller?  I don't think it's a big deal if the controller is over 100 lines, but it would be good to break out the functionality into logical chunks as it continues to grow.

tpendragon: Yeah, may as well. Not sure how we're going to refactor it...maybe a new controller? *Shrug*

tpendragon: #357

escowles: Yeah, I didn't see any big chunks at first glance that seemed like they should be broken out.  But I figure it's worth thinking about.

PR #358, opened 2016-01-19
Sammelband Manifests

tpendragon:
Implements manifests for resources which have both canvases (IE cover pages) AND child scanned resources. This allows MVWs to be Sammelbands. Whether we want to make a Sammelband model is now up to if we need a new workflow or not.

### DIFF
--- a/app/presenters/multi_volume_work_show_presenter.rb
+++ b/app/presenters/multi_volume_work_show_presenter.rb
@@ -1,9 +1,20 @@
 class MultiVolumeWorkShowPresenter < CurationConcernsShowPresenter
   def file_presenter_class
-    ::ScannedResourceShowPresenter
+    DynamicShowPresenter.new
   end
 
   def viewing_hint
     'multi-part'
+  end
+
+  class DynamicShowPresenter
+    def new(*args)
+      solr_doc = args.first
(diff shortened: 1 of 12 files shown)


app/presenters/multi_volume_work_show_presenter.rb
```diff
@@ -1,9 +1,20 @@
...
-    ::ScannedResourceShowPresenter
+    DynamicShowPresenter.new
   end
 
   def viewing_hint
     'multi-part'
   end
+
+  class DynamicShowPresenter
+    def new(*args)
```
escowles: Is there any reason why the body of this method can't just go in `file_presenter_class` above?

tpendragon: Yeah - the branch is based on the argument given to #new on initialization, which isn't available in `file_presenter_class`

escowles: OK, that makes sense.  Then this looks ready to merge -- can you squash the commits?

tpendragon:
@escowles Squashed the two range commits, but I like the separation of the range implementation and the canvas implementation.

jpstroop:
It's not clear to me whether we need a new workflow or not. Could we walk through it tomorrow?

PR #366, opened 2016-01-22
MVW Structures

tpendragon:
Enables structuring a MVW and Sammelband and converting those structures into a IIIF manifest.


app/views/curation_concerns/multi_volume_works/_members.html.erb
```diff
@@ -15,7 +15,11 @@
           <tr>
             <td class="col-xs-3"><%= render_thumbnail_tag res %></td>
             <td>
-              <%= link_to res.title, main_app.curation_concerns_member_scanned_resource_path(@presenter.id, res.id) %>
+              <% if @presenter.model_name.singular == "scanned_resource" %>
```
escowles: Is there a better way to check if it's a ScannedResource?  Not sure if checking the presenter class is any better...

tpendragon: Yeah, this bothered me too. I'm not sure really - I could rescue from an exception if the path doesn't exist?

escowles: Or maybe the presenters could have a method for link to me in the context of X MVW?

tpendragon: Ah, yeah, while writing tests I realized this doesn't even work.

Issue #74, opened 2015-09-14
Accumulate rights statements


jpstroop:
When we're a little further along, show major stakeholders how rights statements / access controls will work and ask them for updated language.

jpstroop:
This is the language that shook out of a recent grant proposal:

> Princeton University Library makes no claims to rights over these images. They are presented for free, on a non-commercial, open-access basis, for fair-use academic and research purposes only. Anyone who claims copyright over any part of these resources and feels that they should not be presented in this manner is invited to contact Princeton University Library, who will in turn consider such concerns and make every effort to respond appropriately.

I think it's a sensible default. @joycebcat, we can have alternatives. Can you think of any major deviations from it from PUDL projects past?

joycebcat:
No, I don’t know of any major deviations from it. —Joyce


On Sep 25, 2015, at 11:14 AM, Jon Stroop <notifications@github.com<mailto:notifications@github.com>> wrote:


This is the language that shook out of a recent grant proposal:

Princeton University Library makes no claims to rights over these images. They are presented for free, on a non-commercial, open-access basis, for fair-use academic and research purposes only. Anyone who claims copyright over any part of these resources and feels that they should not be presented in this manner is invited to contact Princeton University Library, who will in turn consider such concerns and make every effort to respond appropriately.

I think it's a sensible default. @joycebcat<https://github.com/joycebcat>, we can have alternatives. Can you think of any major deviations from it from PUDL projects past?

jpstroop:
First pass:
 * Rights URI for 2.1, 2.3, 2.4, 2.5, 2.7, & 2.9
 * PUL Boilerplate & Citation info
 * Location / Contact INFO
 * Qualifying note

jpstroop:
Initial boilerplate: 

> Princeton University Library claims no copyright governing this digital resource. They are presented for free, on a non-commercial, open-access basis, for fair-use academic and research purposes only. Anyone who claims copyright over any part of these resources and feels that they should not be presented in this manner is invited to contact Princeton University Library, who will in turn consider such concerns and make every effort to respond appropriately.
[please cite us / Attribution]

escowles:
Making more specific tickets #388 and #389, and closing this omnibus ticket.

PR #380, opened 2016-01-26
Adding RightsStatements.org rights statements using QA

escowles:
First part of #74 -- using RightsStatements.org rights statements, and displaying the associated public-facing description, plus generic Princeton boilerplate.


tpendragon:
:+1: Looks and works great!

PR #398, opened 2016-01-29
Set OCR Language

tpendragon:
Gives a place to set a language for OCR and automatically rebuilds fileset OCR if the property changes.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,8 @@ GIT
 
 GIT
   remote: git://github.com/projecthydra-labs/browse-everything.git
-  revision: 7f547a23d8fc20bd249b8f607e009b03aa112c5d
+  revision: 3c3f60472c2b31a3d5a07e8f2bf50f09f8220ed5
+  branch: bootstrap-sprockets
   specs:
     browse-everything (0.9.1)
       bootstrap-sass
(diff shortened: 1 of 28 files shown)


app/models/concerns/ocr.rb
```diff
@@ -0,0 +1,5 @@
+module OCR
+  def ocr?
```
escowles: Is this being used anywhere?

app/views/curation_concerns/base/_iiif_fields.html.erb
```diff
@@ -27,5 +27,9 @@
         </div>
       <% end %>
     </div>
+
+    <div>
+      <%= f.input :ocr_language, collection: [:eng, :ita], input_html: { name: "#{f.object.model_name.singular}[ocr_language][]", multiple: true, class: 'resource-radio-button' } %>
```
escowles: Can we pull the list of languages from the locale info?

config/locales/en.yml
```diff
@@ -30,7 +30,14 @@ en:
 
       Princeton users with copyright questions or concerns should refer to <a href="http://www.princeton.edu/copyright/">http://www.princeton.edu/copyright/</a>.
   simple_form:
+    labels:
+      defaults:
+        ocr_language: "OCR Language"
     options:
+      defaults:
+        ocr_language:
+          eng: "English"
```
escowles: Do we want to include the other languages we expect to handle?  And do we want to include more info (e.g., the ISO 639-2 codes where different)?  This might let us use a single language field for both the language of the resource and the OCR language to use.

Gemfile.lock
```diff
@@ -325,34 +326,26 @@ GEM
...
-      autoparse (~> 0.3)
-      extlib (~> 0.9)
-      faraday (~> 0.9)
-      googleauth (~> 0.3)
-      launchy (~> 2.4)
-      multi_json (~> 1.10)
-      retriable (~> 1.4)
-      signet (~> 0.6)
-    google_drive (1.0.2)
+    google-api-client (0.7.1)
```
escowles: Interesting that the google-api-client is downgraded from 0.8.6 to 0.7.1.

app/models/file_set.rb
```diff
@@ -52,14 +45,10 @@ def ocr_file
...
-      @ocr_text ||=
-        begin
-          if persisted? && File.exist?(ocr_file.gsub("file:", ""))
-            file = File.open(ocr_file.gsub("file:", ""))
-            ocr_doc = HOCRDocument.new(file)
-            ocr_doc.text.strip
-          end
-        end
+      return unless persisted? && File.exist?(ocr_file.gsub("file:", ""))
+      file = File.open(ocr_file.gsub("file:", ""))
```
escowles: Do we need to remove some slashes in addition to `file:`, e.g., `file:///foo/bar` to `/foo/bar`?

app/models/concerns/ocr.rb
tpendragon: Nope, remnant of a refactor.

app/models/file_set.rb
tpendragon: I thought so, but I think the path method actually didn't add enough of them, so this is what was needed.

app/views/curation_concerns/base/_iiif_fields.html.erb
tpendragon: Contemplating shelling out so tesseract can list available languages. I'll give it a shot.

config/locales/en.yml
tpendragon: I was intending the latter, but open to discussion.

escowles: I'm fine with that, esp. to get this merged today.  I added a question ticket to discuss this more: #399

escowles:
Looks good -- squash?

tpendragon:
> Looks good -- squash?

Yeah. Adding support for passing multiple languages to Tesseract real quick too

tpendragon:
@escowles Good?

PR #409, opened 2016-02-09
PDF Cover Page

### LINKED ISSUE #73: PDFs need a boilerplate coverpage
Mudd has this: http://findingaids.princeton.edu/AC044/c0095.pdf

We can probably use the rights statement to trigger different behaviors. We should not go into production w/ the PDF feature until this is sorted out.

tpendragon:
Closes #73

Not the prettiest thing in the world, but it has the necessary data.

### DIFF
--- a/app/renderers/holding_location_renderer.rb
+++ b/app/renderers/holding_location_renderer.rb
@@ -3,6 +3,12 @@ class HoldingLocationRenderer < CurationConcerns::AttributeRenderer
     super(:location, value, options)
   end
 
+  def value_html
+    Array(values).map do |value|
+      attribute_value_to_html(value)
+    end.join("")
+  end
+
   private
 
     def attribute_value_to_html(value)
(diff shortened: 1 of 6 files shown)


tpendragon:
Example PDF: [fixture_pdf.pdf](https://github.com/pulibrary/plum/files/123481/fixture_pdf.pdf)

escowles:
Can you add the title and source metadata id too?  I think those would round out the necessary cover page info.

tpendragon:
> source metadata id too

Is this useful for a PDF? I'd think better would be embedding metadata, which I think has a ticket

escowles:
I think it's nice to have an ID on the cover page that people could use to uniquely identify the thing they were talking about (e.g., if they want to report a problem, or find it again).  The URI might be better than the source metadata ID.

tpendragon:
> The URI might be better than the source metadata ID.

:+1: to that. @jpstroop?

tpendragon:
There's the title/URI.

tpendragon:
Example: 
[fixture_pdf.pdf](https://github.com/pulibrary/plum/files/123502/fixture_pdf.pdf)

escowles:
:+1: looks good

PR #417, opened 2016-02-13
**DO NOT MERGE** Use Curation Concerns File Manager

tpendragon:
Mostly just up so I don't forget. We should be able to merge this once CC has the file manager.


tpendragon:
Mergable when https://github.com/projecthydra-labs/curation_concerns/pull/612 is merged.

tpendragon:
@escowles @jpstroop This should be ready now - file manager is officially in CC 0.8!

escowles:
I was expecting this PR to be almost all deletes, but there are a significant amount of additions, esp. in javascript and views.  Is this because we're adding features on top of the stock CC file manager?  Or is this what you'd expect most implementer to do?

tpendragon:
@escowles It's because we're adding features - mostly the IIIF radio buttons and OCR text selector.

PR #431, opened 2016-02-24
Hide label/legend in search bar.

### LINKED ISSUE #423: Hide "Search Plum: A Repository"

tpendragon:
Closes #423

### DIFF
--- a/app/assets/stylesheets/abstractions/mixins.scss
+++ b/app/assets/stylesheets/abstractions/mixins.scss
@@ -5,3 +5,16 @@
     -ms-transform: translateY(-50%);
     transform: translateY(-50%);
 }
+
+// Hide content visually. still visible to screen readers
+// @include visually-hidden;
+
+@mixin visually-hidden {
+  position: absolute;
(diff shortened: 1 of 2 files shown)


escowles:
Would it be better to override the view than just hide the text with CSS?

jpstroop:
I was wondering the same thing but forgot to comment last night.

axamei:
Screen readers use the label element to describe the related input field so we want to keep that markup. And, setting the label to `display:none;` hides it from the screen reader. 

What we want to do is visually hide it, while making it readable by screen readers. See:  https://github.com/pulibrary/pul_assets/blob/6bb97a16ba5c03d1291c567a151a9e08b8a5d0c1/app/assets/stylesheets/utils/_mixins.scss#L25-L31

tpendragon:
@axamei That seems right. It had that class and didn't seem to work. I'll take a closer look, thanks for the comment!

tpendragon:
@axamei @jpstroop @escowles Updated.

PR #438, opened 2016-02-26
Run at new PDF Rendering.

tpendragon:
Attempt at a new PDF Format


tpendragon:
Looks like the following right now:

[fixture_pdf.pdf](https://github.com/pulibrary/plum/files/148665/fixture_pdf.pdf)

escowles:
This is looking very nice!  The only suggestion I have is underlining the URL.

jpstroop:
I think the logo and labels could be smaller, but agree, it seems like we're on the right path.

jpstroop:
This sample from the Beinecke is interesting, though there's less info than their group wants.

I sort of like the padding around the image as well, but that's probably a separate conversation.

[Beinecke_DL_2016480_1010185.pdf](https://github.com/pulibrary/plum/files/148674/Beinecke_DL_2016480_1010185.pdf)

tpendragon:
Add address to holding location, and add metadata.

PR #445, opened 2016-03-01
Adding coveralls webhook

escowles:


tpendragon:
Doesn't look like Coveralls is reporting?

escowles:
Yeah, Coveralls processed the commit, but didn't report on the PR: https://coveralls.io/builds/5260893

escowles:
Looks like Coveralls and Circle are both happy now!

PR #464, opened 2016-03-03
Use IIIF for FileSet Thumbnails.

### LINKED ISSUE #460: Use image server for thumbnails

tpendragon:
Closes #460

### DIFF
--- a/app/values/iiif_path.rb
+++ b/app/values/iiif_path.rb
@@ -9,6 +9,10 @@ class IIIFPath
     "#{iiif_url}/#{escaped_path}"
   end
 
+  def thumbnail
+    "#{self}/full/!200,150/0/default.jpg"
+  end
+
   private
 
     def escaped_path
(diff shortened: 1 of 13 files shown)


app/values/iiif_path.rb
```diff
@@ -9,6 +9,10 @@ def to_s
     "#{iiif_url}/#{escaped_path}"
   end
 
+  def thumbnail
+    "#{self}/full/100,/0/default.jpg"
```
escowles: Shouldn't this be 200 or 200,150 like we're currently using?

https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/file_set/derivatives.rb#L40-L41

jpstroop: `w,` is the [Canonical Syntax](http://iiif.io/api/image/2.1/#canonical-uri-syntax). It's a minor thing since our server supports all of the size options, but this is probably best for interop/code reuse.

escowles: To replicate the behavior we have now, we would want `!200,150` -- but I would also be fine with `200,` to be more compatible with other image servers.

jpstroop: Yeah. If we have UI constraints that need the image to be at or smaller that 200 x 150, that's the way to go. For portrait images, thanks to some cache magic, Loris is going to serve the same file for `w,` and `!w,h` anyway.

tpendragon: Taken care of.

PR #479, opened 2016-03-08
Update solr_wrapper, use more of the ActiveFedora Solr/Fedora wrappers

escowles:
* Removes jetty_wrapper
* Brings our Rake tasks more in line with CurationConcerns, etc.
* `rake spec` runs style checkers and tests, `rake ci` spins up Solr/Fedora and runs them


escowles:
I improved the rake `ci` task to run `rubocop` & `eslint`.  I tried to include `spec:javascript`, but it made the task shell out and then hang afterwards.  So I think we will need Circle to keep doing `server:test` and `jetty_wait` until that can be resolved.

tpendragon:
@escowles :+1:! I think we'll always want Circle to do server:test and jetty_wait. If Circle can guess what command to use, then it's able to automatically split tests between parallel builds.

PR #480, opened 2016-03-08
Using resource language as default OCR language - closes #399

### LINKED ISSUE #399: Set the OCR language default based on the language of the resource
E.g., if the language of the resource is Latin, the OCR should default to Italian.

escowles:


tpendragon:
My only concern here is there's a good chance the language from the language field isn't an available language for Tesseract.

escowles:
OK, I've updated the logic to check to make sure language is in the Tesseract's language list.  Also had to tweak the existing test to avoid the cached list overriding the stubbed language list.

PR #485, opened 2016-03-10
Add navigation date to form.

### LINKED ISSUE #284: Add navDate to create/edit form
See last property here: http://iiif.io/api/presentation/2.1/#technical-properties

tpendragon:
Closes #284

Some questions first: I left the time selectors turned on, but it seems unlikely time would ACTUALLY be entered, yes? Maybe I should turn off the selector and make it always 00:00:00 somehow?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,4 +107,5 @@ gem 'bunny'
 source 'https://rails-assets.org' do
   gem 'rails-assets-babel-polyfill'
   gem 'rails-assets-bootstrap-select', '1.9.4'
+  gem 'rails-assets-jqueryui-timepicker-addon'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -557,6 +557,7 @@ GEM
     rails-assets-bootstrap-select (1.9.4)
       rails-assets-jquery (>= 1.8)
     rails-assets-jquery (2.2.0)
+    rails-assets-jqueryui-timepicker-addon (1.6.1)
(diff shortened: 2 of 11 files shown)


escowles:
Agree entering time seems pretty unlikely.  How much does it get in the way?  Can the user just ignore the time or do they have to pick something?

escowles:
I could go either way: hiding the time means less clutter and will probably never be a problem (and we can always turn it back on later).  On the other hand, if they don't have to interact, it seems fine to show it and it could hypothetically be known, or used to order multiple items from the same date.

escowles:
The UTC timezone gets trickier if the time is shown (so that's another reason to hide it).  If the time is hidden, then I think we don't need to worry as much about localtime/UTC being off by a day.

tpendragon:
This may just go to cataloguing practices, so it may not be an issue. When someone says it was created on the 19th, is time zone taken into account?

tpendragon:
I think I'm gonna hide the time controls and we can turn them on later.

jpstroop:
> This may just go to cataloguing practices, so it may not be an issue. When someone says it was created on the 19th, is time zone taken into account?

Let's wait for a use case. For now, I think stopping at the day is sufficient.

jpstroop:
The intent is to be able to, e.g., sort serials put them on a calendar, etc. I suppose there could be morning and evening editions of newspapers, but I don't think we have that use case, at least yet. Better to just zero out the time I think.

PR #487, opened 2016-03-11
Implementing IIIF token auth

### LINKED ISSUE #466: Implement IIIF Token Auth

escowles:
Closes #466

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,9 +19,23 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  def current_user
+    return super unless super.nil?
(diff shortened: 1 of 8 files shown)


tpendragon:
I might be missing something here. Does this actually auth a user via CAS and give them a token to maintain that auth? It looks more like it'll give a token to whoever asks and treat them as if they're on campus?

escowles:
Oh, right -- I do need to make sure they're authenticated before I give them the token...

tpendragon:
:+1: I think the token needs to be associated with the user they're authenticated as too.

escowles:
The token now includes the user's email in addition to the expiration timestamp.  So the controller uses the email to lookup their user object.

tpendragon:
Trying this out on UV, might take me a bit.

escowles:
Sounds good -- definitely want to make sure it really works.

PR #496, opened 2016-03-15
Fix MVW rendering in Collection Manifests.

tpendragon:


app/presenters/collection_show_presenter.rb
```diff
@@ -31,8 +31,7 @@ def ordered_ids
       Array(solr_document["member_ids_ssim"])
     end
 
-    # Override this method if you want to use an alternate presenter class for the files
     def file_presenter_class
-      ScannedResourceShowPresenter
+      MultiVolumeWorkShowPresenter::DynamicShowPresenter.new
```
escowles: This is returning an instance instead of the class -- is that OK?

tpendragon: Yeah, it's necessary.

Issue #493, opened 2016-03-14
MultiVolumeWork collections not selected in edit form


escowles:
When I edit a MultiVolumeWork that is in a collection, the edit form does not have the collection pre-selected.  If I submit the form without re-selecting the collection, it is removed from the collection.

escowles:
Something weird is going on here -- it looks like the MVW form is using the MultiVolumeWorkForm, but ScannedResource is using the record itself.

Issue #494, opened 2016-03-14
Default rights statement for MultiVolumeWorks


escowles:
When creating a MultiVolumeWork, the No Known Copyright rights statement should be selected by default, like it is with ScannedResources.

tpendragon:
This is related to #493, and I think I have the fix. Running tests now.

PR #500, opened 2016-03-15
Use form for Scanned Resources.

tpendragon:


escowles:
This should fix #493 and #494, right?  Do you want to add some feature tests for those?

tpendragon:
That code actually got edited because SR tests broke without them. It DOES fix #493 and #494, but only because they both act the same now. Should we do both tests?

escowles:
I'm torn -- in general I think we don't want to duplicate all the tests for SR and MVW where the behavior is the same, so I'm fine merging this without adding extra tests.  But it was different and we didn't notice because there isn't anything in the tests to make sure MVW is behaving the same.

PR #502, opened 2016-03-16
Initial METS ingest implementation

escowles:


escowles:
Here is a sample object ingested with this rake task: https://hydra-dev.princeton.edu/concern/scanned_resources/3t945q79m

tpendragon:
@escowles Is labeling the files that are ingested going to happen in another PR?

jpstroop:
The images don't match the object: http://pudl.princeton.edu/sheetreader.php?obj=d217qp581 cache problem again?

escowles:
I've updated to use the FileSetActor, which handles the initial file labels and derivatives correctly.  Updated example: https://hydra-dev.princeton.edu/concern/scanned_resources/05741r719

PR #517, opened 2016-03-31
Prevent phantomjs auto-install (occasional failures causing CI failures)

escowles:


escowles:
Copying this [from Sufia](https://github.com/projecthydra/sufia/pull/1712/files) since I've seen some CircleCI failures because the PhantomJS download fails.  I assumed it was needed to run the jasmine tests, but the build is passing without it.

PR #526, opened 2016-04-19
Hide records when not complete or flagged.

### LINKED ISSUE #525: Enforce workflow visibility

tpendragon:
Closes #525.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -1,7 +1,11 @@
 class SearchBuilder < CurationConcerns::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
+  delegate :unreadable_states, to: :current_ability
 
-  self.default_processor_chain += [:hide_parented_resources, :join_from_parent]
+  self.default_processor_chain += [
+    :hide_parented_resources, :join_from_parent,
+    :hide_incomplete
+  ]
 
   def self.show_actions
     [:show, :manifest, :structure, :pdf]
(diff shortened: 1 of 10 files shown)


app/models/search_builder.rb
```diff
@@ -18,6 +22,13 @@ def join_from_parent(solr_params)
     solr_params[:q] = JoinChildrenQuery.new(solr_params[:q]).to_s
   end
 
+  def hide_incomplete(solr_params)
+    return if current_user.admin? || current_user.editor? || current_user.curator? || current_user.fulfiller? || current_user.image_editor?
+    solr_params[:fq] ||= []
+    state_field = ActiveFedora.index_field_mapper.solr_name('state', :symbol)
+    solr_params[:fq] << "#{state_field}:complete OR #{state_field}:flagged"
```
escowles: Can any this or the admin/editor/etc. checks be moved into Ability?  Even if it winds up being the same logic, it would be better to have that in the same place as the existing permissions.

tpendragon: I really wanted to use Ability to do this bit, but it requires a record to check against which has a persisted ID (because of Blacklight::AccessControls). I could move this method over there, but I don't think `ability.rb` would use it, since it uses CanCan's DSL.

tpendragon: Maybe I can use the block syntax in CanCan...

escowles: If nothing else, you could add methods to Ability to test whether the current user can view everything, and list the states anonymous users can view.  Having that logic in Ability instead of the SearchBuilder would make it easier to maintain.

tpendragon: I think I have the two sharing logic now, running specs and then I'll push it up.

tpendragon:
@escowles Ready for another look.

PR #529, opened 2016-04-20
Suggest portion note for records with duplicate source metadata identifiers

### LINKED ISSUE #469: Suggest using a portion note when there are duplicate bib IDs
Follow on from #207. It seems like some indication of why this is different should be a requirement

escowles:
Closes #469

### DIFF
--- a/app/assets/javascripts/detect_duplicates.js
+++ b/app/assets/javascripts/detect_duplicates.js
@@ -10,7 +10,7 @@ function detect_duplicates() {
     $.ajax({ url: url + '&format=json' })
     .done(function(data) {
       if ( data.docs.length > 0 ) {
-        $me.parent().append('<div id="duplicates" class="alert alert-warning">This ID is already in use: <a href="' + url + '">view records using this Source Metadata ID</a>.</div>');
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/detect_duplicates.js
```diff
@@ -10,7 +10,7 @@ function detect_duplicates() {
     $.ajax({ url: url + '&format=json' })
     .done(function(data) {
       if ( data.docs.length > 0 ) {
-        $me.parent().append('<div id="duplicates" class="alert alert-warning">This ID is already in use: <a href="' + url + '">view records using this Source Metadata ID</a>.</div>');
+        $me.parent().append('<div id="duplicates" class="alert alert-warning">This ID is already in use: <a href="' + url + '">view records using this Source Metadata ID</a>.  Please use a portion note to differentiate records with the same ID.</div>');
```
jpstroop: Maybe "Please consider using the Portion Note field to help differentiate between objects with the same metadata" ?

escowles:
Sounds like a better message -- PR updated.

PR #530, opened 2016-04-20
Enable RTL rendering.

tpendragon:


Rakefile
```diff
@@ -34,11 +34,7 @@ namespace :server do
   desc 'Run Fedora and Solr for development environment'
   task :development do
     run_server 'development', solr_port: 8983, fcrepo_port: 8984 do
-      IO.popen('rails server') do |io|
-        io.each do |line|
-          puts line
-        end
-      end
+      sleep
```
tpendragon: Ergh. I can back this out, but I really do prefer it. Shutting down and losing solr records every time you need to reload server config is a huge bother.

escowles: We should remove that anyway, now that it has mostly been replaced by `rake hydra:server` -- I'd be fine with keeping a version that doesn't run rails in Plum and/or adding it to hydra-head (as `rake hydra:dev_server` maybe?).  See https://github.com/projecthydra/hydra-head/blob/master/hydra-core/lib/tasks/hydra.rake.  Either way, should be a separate PR.

PR #543, opened 2016-04-25
Color pdf toggle

### LINKED ISSUE #511: PDF Options
We need two flavors of PDF: Gray and Color, and an option to make a PDF not available at all.

Use case for Gray:
Smaller file, degraded quality w/ compression and 8 instead of 24 bit depth serves the reading use case while limiting it's use as a reproduction

Use case for Color:
Larger, but truer fidelity to the RWO, useful for printing 'preservation photocopies'

Use case for None:
Concerned curators.

Discussion on 3/29 suggested a we find a predicate to indicate the download availability, and URIs for `:true`, `:false`, and `:degraded`.

tpendragon:
We should probably figure out the predicates/objects for this, but want to make sure this is the necessary functionality.

Effectively this adds a dropdown to the form to select grayscale or color. It defaults to grayscale. If you select the empty option, then there's no PDF available.

Closes #511.

### DIFF
--- a/app/controllers/curation_concerns/scanned_resources_controller.rb
+++ b/app/controllers/curation_concerns/scanned_resources_controller.rb
@@ -33,8 +33,20 @@ class CurationConcerns::ScannedResourcesController < CurationConcerns::CurationC
   private
 
     def authorize_pdf
-      return unless params[:pdf_quality] == "color"
-      return if can?(:color_pdf, ScannedResource)
+      if params[:pdf_quality] == "color"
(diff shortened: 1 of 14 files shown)


jpstroop:
I think this if fine. We can refine if necessary. Most important to demo what it possible (and that no PDF is possible) for now.

escowles:
+1 — not crazy about pdf_type property, but the functionality is what we need, so we can worry about the predicates later.

PR #583, opened 2016-05-13
Restore styles removed from cc update

axamei:


tpendragon:
You're the best. I'll merge when this passes.

PR #601, opened 2016-05-17
Allow providing a list of collections that migrated records should be members of

### LINKED ISSUE #587: METS migration should handle collection membership

escowles:
Closes #587

### DIFF
--- a/app/jobs/ingest_mets_job.rb
+++ b/app/jobs/ingest_mets_job.rb
@@ -3,10 +3,12 @@ class IngestMETSJob < ActiveJob::Base
 
   # @param [String] mets_file Filename of a METS file to ingest
   # @param [String] user User to ingest as
-  def perform(mets_file, user)
+  # @param [Array<String>] collections Collection IDs the resources should be members of
+  def perform(mets_file, user, collections = [])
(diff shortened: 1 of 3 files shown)


app/jobs/ingest_mets_job.rb
```diff
@@ -3,10 +3,12 @@ class IngestMETSJob < ActiveJob::Base
 
   # @param [String] mets_file Filename of a METS file to ingest
   # @param [String] user User to ingest as
-  def perform(mets_file, user)
+  # @param [Array<String>] collections Collection IDs the resources should be members of
+  def perform(mets_file, user, collections = [])
     logger.info "Ingesting METS #{mets_file}"
     @mets = METSDocument.new mets_file
     @user = user
+    @collections = collections
```
tpendragon: Maybe you should find the collection objects here, so it'll throw the error early if they're bad IDs?

escowles: That's a good idea.  I dithered about putting that in the rake task, but decided I wanted to keep it as simple as possible.  But finding them here would prevent ingesting records before a failure.

PR #605, opened 2016-05-18
Update Curation Concerns

### LINKED ISSUE #579: When scrolling, the toolbar header should scroll past in file manager.
Allows for more vertical space.

tpendragon:
Closes #579.

### DIFF
--- a/.fcrepo_wrapper
+++ b/.fcrepo_wrapper
@@ -1,0 +1,4 @@
+# Place any default configuration for solr_wrapper here
+port: 8984
+enable_jms: false
+fcrepo_home_dir: tmp/fcrepo4-development-data
--- a/.solr_wrapper
+++ b/.solr_wrapper
@@ -1,0 +1,8 @@
+# Place any default configuration for solr_wrapper here
+# version: 6.0.0
+# port: 8983
+instance_dir: tmp/solr-development
+collection:
+  persist: true
+  dir: solr/config/
+  name: hydra-development
--- a/Gemfile
+++ b/Gemfile
@@ -66,8 +66,8 @@ group :development, :test do
   gem 'pry-rails'
 end
 
-gem 'solr_wrapper', '~> 0.5.1'
-gem 'fcrepo_wrapper', '~> 0.2'
+gem 'solr_wrapper', '~> 0.12.0'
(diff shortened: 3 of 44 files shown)


app/presenters/collection_show_presenter.rb
```diff
@@ -1,6 +1,6 @@
 class CollectionShowPresenter < CurationConcerns::CollectionPresenter
   include CurationConcerns::Serializers
-  delegate :id, :title, :exhibit_id, :date_created, to: :solr_document, allow_nil: true
+  delegate :id, :title, :exhibit_id, :date_created, :keyword, to: :solr_document, allow_nil: true
```
escowles: Are we adding keywords?

tpendragon: CC changed "tag". https://github.com/projecthydra/curation_concerns/commit/442633637cde9d8c82b7f0df28cf28493f494b37

escowles: That seems reasonable, but we're not providing UI to edit them now — do we need to include keywords in the presenter, or can we just leave it out like we did when it was tags?

tpendragon: I'm not sure where it's coming from, but I added it because it was throwing test errors. I can try again.

app/views/curation_concerns/base/file_manager.html.erb
```diff
@@ -0,0 +1,34 @@
+<h1><%= t("file_manager.link_text") %></h1>
```
escowles: Is the same as the CC version?  I just eyeballed them and didn't see any difference.

tpendragon: Yup! Thanks, it was there when I was debugging something.

escowles:
Looks like `rake server:test` isn't working because the update solr_wrapper/fcrepo_wrapper and rake task updates.  I'll update the Rakefile and add a commit  to this branch.

tpendragon:
@escowles Thank you!

escowles:
@tpendragon the CC upgrade looks good to me — go ahead and merge this if you're OK with the fcrepo/solr_wrapper updates.

PR #611, opened 2016-05-19
Better collapse functionality for Structure

tpendragon:
Now dragging into a collapsed structure won't expand it, and dragging
right of a collapsed structure creates a new node.

### DIFF
--- a/app/assets/javascripts/structure_editor.es6
+++ b/app/assets/javascripts/structure_editor.es6
@@ -27,6 +27,7 @@ jQuery(() => {
     listType: "ul",
     placeholder: "placeholder",
     parentNodeFactory: new_node,
+    preventExpansion: true,
     helper: function(e, item) {
       if ( ! item.hasClass("ui-selected") ) {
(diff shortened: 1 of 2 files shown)


tpendragon:
https://vimeo.com/167344323 Demo of the new process. @jpstroop @joycebcat ?

jpstroop:
That looks better to me. So a structure has to be collapsed in order to create a sibling? I think that makes sense.

tpendragon:
@jpstroop It doesn't -have- to, you can make one the same way you used to - drag too far and then drag left. You just -can- do it this way now.

joycebcat:
@tpendragon I am unable to replicate what I see in the video.  Is it in hydra-dev now?  Everytime I try to create a new section as shown, the collapsed section pops open and either nothing new is created or they end up nested under the first section.  I find that I have sometimes actually changed the order of the images while I'm trying to create a section where I want to place it.

tpendragon:
No, it won't be on hydra dev until this PR is merged. Sorry for the confusion!

PR #612, opened 2016-05-20
Updating to Solr 6

escowles:
* Using Solr 6, and updating Solr config and query generation to work with Solr 6


escowles:
Do we want to start over with a fresh repo, or do we want to reindex everything in Fedora?

tpendragon:
Let's just do a fresh one. Old items have bad ACLs anyways

PR #623, opened 2016-05-23
Updating display of collection description and title

### LINKED ISSUE #619: Collection display issues
On the collection show page, the description is showing up three times (twice below the viewer, a third time in the attributes table).  It should show up only in the attributes table.

On the item show page, collections titles include brackets and quotes, which should be removed.

escowles:
Fixes #619

### DIFF
--- a/app/views/collections/show.html.erb
+++ b/app/views/collections/show.html.erb
@@ -1,0 +1,30 @@
+<% provide :page_title, "#{@presenter.title.join(', ')} - #{application_name}" %>
+
+<% provide :sidebar do %>
+  <%= render 'search_sidebar' %>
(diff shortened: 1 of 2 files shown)


app/views/collections/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% provide :page_title, "#{@presenter.title.join(', ')} - #{application_name}" %>
```
tpendragon: `presenter.title.to_sentence`?

tpendragon: Eh, maybe not. An "and" probably isn't appropriate for multiple titles.

PR #657, opened 2016-06-06
Update Curation Concerns

tpendragon:
Uses CC's nested works.

Closes #657


app/jobs/ingest_file_job.rb
```diff
@@ -2,31 +2,39 @@ class IngestFileJob < ActiveJob::Base
...
 
+    repository_file = file_set.send(relation.to_sym)
+
     # Do post file ingest actions
-    CurationConcerns::VersioningService.create(file_set.send(relation.to_sym), user)
-    CharacterizeJob.perform_later(file_set, filename)
+    CurationConcerns::VersioningService.create(repository_file, user)
+
+    # TODO: this is a problem, the file may not be available at this path on another machine.
+    # It may be local, or it may be in s3
```
escowles: I think this needs a general solution, so I've created https://github.com/pulibrary/plum/issues/660 for looking into that.

tpendragon: I should note, this wasn't my TODO. This is just copied from CC.

escowles: OK, that's good to know — I hope we'll be able to update CC (or hydra_derivs) to pull files from Fedora if they're not on local disk.

PR #668, opened 2016-06-08
Reuse OSD Viewer

### LINKED ISSUE #654: OSd zoom on 2nd and greater image is jerky
The zoom is always smooth, after that the the zoom into addition thumbnails is jerky.  The message

> Hash Openseadragon1 has already been used.

appears in the JS console for each additional zoom. Might be related?

tpendragon:
Closes #654 I think.

@jpstroop this is deployed to staging now, when you get a chance can you confirm that it's better?


jpstroop:
:+1:

PR #680, opened 2016-06-28
Push touching parent works to the background

tpendragon:


Gemfile
```diff
@@ -108,6 +108,7 @@ gem 'string_rtl'
 gem 'sinatra'
 gem 'redis-namespace'
 gem 'arabic-letter-connector'
+gem 'ruby-prof'
```
escowles: Do we want to take ruby-prof out, or limit to development/staging envs?

tpendragon:
@escowles Agreed on restricting environment - added.

PR #681, opened 2016-06-29
Right-aligning RTL text, adding fonts, correcting lam alef ligature

escowles:
* Right-aligning RTL text
* Correcting lam alef ligature

[arabic-cover.pdf](https://github.com/pulibrary/plum/files/340218/arabic-cover.pdf)


escowles:
@joycebcat I think this resolves most of the issues you raised, except the low position of the macrons.

escowles:
@khaledhosny I don't know if you've worked with the Prawn PDF-generation library, but I wasn't able to get the lam alef ligature to display correctly unless I explicitly replaced the characters (see https://github.com/pulibrary/plum/pull/681/files#diff-20820bb1c996d33b862d41e82a38690fR96).  I'd appreciate any advice you can give, and of course feedback on whether the Arabic text is improved in general.

khaledhosny:
I haven't used before, but I guess it is doing rudimentary presentation forms shaping which works only for the simplest of texts and fonts and small set of languages that has Arabic presentation forms in Unicode.

For proper Arabic shaping (and other scripts like Indic and South Asian scripts), you need to use an OpenType layout engine like https://github.com/behdad/harfbuzz to do the low level text layout.

PR #707, opened 2016-08-23
Remove the attach a file button from scanned resources show page

### LINKED ISSUE #669: Remove Attach a File

eliotjordan:
Removes the `Attach a File` button from the scanned resources show page. Closes #669.

### DIFF
--- a/spec/features/edit_scanned_resource_spec.rb
+++ b/spec/features/edit_scanned_resource_spec.rb
@@ -45,18 +45,17 @@ RSpec.feature "ScannedResourcesController", type: :feature do
       expect(page).to have_text("Error retrieving metadata")
     end
 
-    scenario "User can add a new file" do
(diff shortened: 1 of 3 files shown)


spec/features/edit_scanned_resource_spec.rb
```diff
@@ -45,27 +45,6 @@
       expect(page).to have_text("Error retrieving metadata")
     end
 
-    scenario "User can add a new file" do
```
escowles: Would it be possible to rework this test to work with the File Manager?  It would be great to keep a test that actually uploaded a file.

eliotjordan: I had the same thought. I'll give it a shot.

eliotjordan: I had the same thought. I'll give it a shot.

PR #712, opened 2016-08-24
Maintain literals when copying from JSONLD

tpendragon:


escowles:
The behavior I see when using this branch, is:
- Titles have language tags (retrieved from bibdata)
- Updates to other fields don't remove them
- Other fields (e.g., creator) do not have langauge tags (bibdata doesn't present them with language tags, see https://bibdata.princeton.edu/bibliographic/4746714/jsonld)

So I'm going to merge this PR and close pulibrary/bibdata#589.  I have created an issue for bibdata to add language tags to non-title fields: https://github.com/pulibrary/marc_liberation/issues/154.  When that's in place, we can revisit whether those language tags are brought in and preserved after editing.

PR #714, opened 2016-08-25
Adjusting fonts and sizing of RTL text

### LINKED ISSUE #632: Font for Arabic
PUDL uses `font-family: 'Times New Roman', 'Arial Unicode MS', serif;`

escowles:
Fixes #632 

Here is what the titles look like before:
plum-before

And here is what they look like after this change:
plum-after


jpstroop:
Looks good!

PR #734, opened 2016-09-14
Prefix environment to NOIDs.

tpendragon:
Fixes name conflicts for the shared image server.


tpendragon:
I suspect this will break anything ingested right now.

escowles:
I'm fine with breaking everything ingested now, to avoid collisions in the future.

config/initializers/curation_concerns.rb
```diff
@@ -34,10 +34,10 @@
   # config.enable_noids = true
 
   # Specify a different template for your repository's NOID IDs
-  # config.noid_template = ".reeddeeddk"
+  config.noid_template = "#{Rails.env.first.downcase}.reeddeeddk"
```
jpstroop: Should we add a `z`?

```
for sequentially generated identifiers, unlimited in length or number, re-using the
most significant mask character (the second character of the Mask) as needed.
```
Then we _know_ we'll never run out.

PR #752, opened 2016-10-12
Always copy file to working directory

tpendragon:
Fixes problems with having worker server different than application
server.


eliotjordan:
:+1:

PR #754, opened 2016-10-12
Use an incrementing ID

### LINKED ISSUE #735: Add a `z` to the NOID template
```
`z` - for sequentially generated identifiers, unlimited in length or number, re-using the
most significant mask character (the second character of the Mask) as needed.
```
Then we _know_ we'll never run out.

tpendragon:
Closes #735.

### DIFF
--- a/config/initializers/curation_concerns.rb
+++ b/config/initializers/curation_concerns.rb
@@ -34,7 +34,7 @@ CurationConcerns.configure do |config|
   # config.enable_noids = true
 
   # Specify a different template for your repository's NOID IDs
-  config.noid_template = "#{Rails.env.first.downcase}.reeddeeddk"
+  config.noid_template = "#{Rails.env.first.downcase}.zeeddeeddk"
 
   # Specify the minter statefile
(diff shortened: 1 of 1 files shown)


config/initializers/curation_concerns.rb
```diff
@@ -34,7 +34,7 @@
   # config.enable_noids = true
 
   # Specify a different template for your repository's NOID IDs
-  config.noid_template = "#{Rails.env.first.downcase}.reeddeeddk"
+  config.noid_template = "#{Rails.env.first.downcase}.zeeddeeddk"
```
tpendragon: > The first letter of the Mask, the generator type, determines the order and boundedness of the namespace. For example, in the Template .sddd, the Prefix is empty and the s says that the namespace is sequentially generated but bounded. The generator type may be one of,

jpstroop: hold on, I'll check the docs.

jpstroop: Yup, you're right.

PR #773, opened 2016-10-18
Do ordering only after all the FileSets are loaded

escowles:
Use a custom FileSetActor to avoid locking the parent work, adding to ordered_members, etc.


tpendragon:
So the thing I think that doesn't work, is if the object fails before ordered_members are assigned I THINK the second try won't delete the filesets attached to the work.

tpendragon:
I think there's a test for that though, so I may be wrong...

escowles:
The FileSets are added to the non-ordered members list — is that not enough for them to be found and destroyed when the parent Work is destroyed?

PR #779, opened 2016-10-19
add back to search links

### LINKED ISSUE #743: Navigation: How to get back to search results?
Am I missing it? Did we take something out of Blacklight? Once I'm on the show page for an object, I can only get back to search results via my back button.

tampakis:
Closes #743

### DIFF
--- a/app/assets/stylesheets/components/catalog.scss
+++ b/app/assets/stylesheets/components/catalog.scss
@@ -15,3 +15,7 @@ dl.document-metadata {
     padding-left: 0;
   }
 }
+
+.back-to-search {
+  margin-bottom: 10px;
+}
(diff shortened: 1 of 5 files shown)


tpendragon:
Isn't the blacklight paging stuff missing really the issue? Although, since it's out of blacklight in the show view maybe this is what we have to do...

tampakis:
This is how Blacklight does the "back to search" link - https://github.com/projectblacklight/blacklight/blob/e7d7cb883e4d24a5263797cac47a8e5fd5b603ca/app/views/catalog/show.html.erb#L1-L6

I couldn't use the `link_back_to_catalog` method though because Rails got confused about which catalog controller to use.

PR #802, opened 2016-11-01
Use the database for minter state

### LINKED ISSUE #801: Use database for Noid minter state

escowles:
Fixes #801


tpendragon:
Does this mean we can delete that minter_state class we have copied over?

escowles:
We can copy the minter state from the state file to the database with the rake task: 
`rake active_fedora:noid:migrate:file_to_database`

tpendragon:
I mean can we delete this: https://github.com/pulibrary/plum/blob/master/app/models/minter_state.rb#L1

escowles:
@tpendragon I think so — let me test it locally to see.

escowles:
Yep, it works fine without `app/models/minter_state.rb` now.

tpendragon:
I think it was a problem deploying before - can you try deploying that to staging?

escowles:
I can try it, but it'll fail because of the plum-workers issue.

escowles:
This build is deployed to staging and working now.  To get the db minter working, I had to clear out an invalid seed before migrating the minter state (see instructions: https://gist.github.com/escowles/c5f1b505954ab368fa659614a2a85610)

escowles:
One more thing before merging this: I changed the development database connect to use Postgres to make it easier to test — I'll change it back to sqlite now.

PR #805, opened 2016-11-04
Setting viewing hit to paged, except for certain types

### LINKED ISSUE #804: METS ingest should default to viewingHint="paged"
See also #803 and the [migration notes](https://docs.google.com/document/d/1yfGaSn_hNXcQILX2LrskuPxTSxK32jKPqNAN8mjDk7I/edit#heading=h.9h3opmgyueht)

escowles:
Fixes #804

### DIFF
--- a/app/models/mets_document.rb
+++ b/app/models/mets_document.rb
@@ -36,6 +36,12 @@ class METSDocument
     @mets.xpath("/mets:mets/mets:structMap[@TYPE='Physical']/mets:div/@TYPE").to_s.start_with? 'RTL'
   end
 
+  def viewing_hint
+    type = @mets.xpath("/mets:mets/mets:structMap[@TYPE='Physical']/mets:div/@TYPE").to_s
+    return if ['TightBoundManuscript', 'ScrollSet', 'BoundArt'].any? { |w| type.include?(w) }
+    'paged'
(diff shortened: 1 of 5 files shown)


app/models/mets_document.rb
```diff
@@ -36,6 +36,12 @@ def right_to_left
     @mets.xpath("/mets:mets/mets:structMap[@TYPE='Physical']/mets:div/@TYPE").to_s.start_with? 'RTL'
   end
 
+  def viewing_hint
+    type = @mets.xpath("/mets:mets/mets:structMap[@TYPE='Physical']/mets:div/@TYPE").to_s
+    return if ['TightBoundManuscript', 'ScrollSet', 'BoundArt'].any? { |w| type.include?(w) }
```
jpstroop: One more here: `RTLTightBoundManuscript`

escowles: Because we're doing `type.include?('TightBoundManuscript')` anything with that substring will pass, so `RTLTightBoundManuscript`, `FooTightBoundManuscript`, `TightBoundManuscriptsAreCool` etc.

jpstroop: Oh, I see. :ok_hand:

PR #850, opened 2016-12-01
Add geo extensions to PairtreeDerivativePath

### LINKED ISSUE #842: Overrides in DownloadBehavior
Plum and GeoConcerns both override `CurationConcerns::DownloadBehavior#load_file`.

- https://github.com/pulibrary/plum/blob/master/app/controllers/downloads_controller.rb#L19

- https://github.com/projecthydra-labs/geo_concerns/blob/master/app/controllers/concerns/geo_concerns/download_behavior.rb#L12

### LINKED ISSUE #844: Plum overrides `derivative_url` method
Plum overrides the the `derivative_url` method in `FileSet` which skips `derivative_path_factory`. GeoConcerns uses `GeoConcerns::DerivativePath`  while Plum uses `PairtreeDerivativePath`.

- https://github.com/pulibrary/plum/blob/master/app/models/file_set.rb#L82

- https://github.com/projecthydra-labs/geo_concerns/blob/master/app/models/concerns/geo_concerns/file_set/derivatives.rb#L67

eliotjordan:
Merges into geo_concerns branch.

Adds geo extensions to PairtreeDerivativePath. Closes #844 Closes #842

### DIFF
--- a/app/values/pairtree_derivative_path.rb
+++ b/app/values/pairtree_derivative_path.rb
@@ -9,19 +9,19 @@ class PairtreeDerivativePath < CurationConcerns::DerivativePath
   end
 
   def extension
-    case destination_name
-    when 'thumbnail'
-      ".#{MIME::Types.type_for('jpg').first.extensions.first}"
-    when "intermediate_file"
-      ".jp2"
-    when "ocr"
-      ".hocr"
-    when "gray-pdf"
-      ".pdf"
-    when "color-pdf"
-      ".pdf"
-    else
-      ".#{destination_name}"
-    end
+    value = extensions[destination_name]
+    value ? value : ".#{destination_name}"
+  end
+
+  def extensions
+    {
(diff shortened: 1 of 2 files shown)


app/values/pairtree_derivative_path.rb
```diff
@@ -9,17 +9,35 @@ def file_name
   end
 
   def extension
+    pdf_extensions || geo_extensions || other_extensions
```
escowles: The list of cases is getting pretty big — and it might grow quite a bit more if we add support for A/V or other kinds of files later.  Maybe having a hash mapping destinations to extensions would be simpler?

eliotjordan: I was worried about that. A hash is a good idea. I'll cook something up...

eliotjordan: Updated to use hash.

PR #854, opened 2016-12-02
Don't ingest files into Fedora

escowles:
* Skip ingesting files into Fedora
* DownloadController serves files from disk instead of Fedora
* Characterization/Derivatives/OCR use files from disk


escowles:
Two things:

* The failing test makes me think there's something still not right with the OCR process.
* The magic Fedora external-content MIME type (which includes the external URL) is overwritten by characterization.  So we probably want to restore it afterwards.

PR #855, opened 2016-12-02
Trigger manifest event in fileset actor only if parent is not a geo work

### LINKED ISSUE #851: Only certain work types should generate a manifest event.
https://github.com/pulibrary/plum/blob/master/app/services/manifest_event_generator.rb#L31

eliotjordan:
Merges into geo_concerns branch.

Adds logic to not fire a manifest record updated event when the parent is a geo work. Geo works aren't synced with pomegranate (yet). Closes #851

### DIFF
--- a/app/actors/file_set_actor.rb
+++ b/app/actors/file_set_actor.rb
@@ -1,7 +1,7 @@
 class FileSetActor < ::CurationConcerns::Actors::FileSetActor
   def attach_file_to_work(*args)
     super.tap do |_result|
-      messenger.record_updated(args.first)
+      messenger.record_updated(args.first) unless args.first.is_a? GeoConcerns::BasicGeoMetadata
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/actors/file_set_actor.rb
```diff
@@ -1,7 +1,8 @@
 class FileSetActor < ::CurationConcerns::Actors::FileSetActor
   def attach_file_to_work(*args)
     super.tap do |_result|
-      messenger.record_updated(args.first)
+      record = args.first
+      messenger.record_updated(record) unless geo_work? record
```
escowles: Can you shorten this to:

```
messenger.record_updated(args.first) unless args.first.is_a? GeoConcerns::BasicMetadata
```

?

eliotjordan: Updated and pushed.

PR #891, opened 2016-12-16
Fix Geoblacklight document generation

### LINKED ISSUE #890: Geoblacklight documents are not being generated properly
- Issue with rails 4 dates in XMLSchema :( This needs to be fixed upstream in GeoConcerns
- layer geometry is not being created
- Need to use different institution name for slug.

eliotjordan:
Closes #890


tampakis:
Looks good, the slug spec is now a lot cleaner too.

PR #901, opened 2016-12-21
Use curation concern base view partials for geo file manager

### LINKED ISSUE #892: Regenerating derivatives on geo filesets throws an error
Something about viewing direction.

tampakis:
Closes #892. Uses the curation concerns gem base partials for the geo-specific models.


eliotjordan:
Needs a rebase against geo_concerns.

PR #933, opened 2017-01-06
Replace aasm with sipity workflow

eliotjordan:
Pull request to add sipity workflows to the CC 2.0 branch.


spec/support/shared_examples/structured_controller.rb
```diff
@@ -1,6 +1,6 @@
 RSpec.shared_examples "structure persister" do |resource_symbol, presenter_factory|
   describe "#structure" do
-    let(:user) { FactoryGirl.create(:user) }
+    let(:user) { FactoryGirl.create(:admin) }
```
eliotjordan: This spec needs a closer look. Definitely some sipity-related permissions issues.

app/models/ability.rb
```diff
@@ -24,7 +24,7 @@ def image_editor_permissions
     can [:file_manager, :save_structure], MultiVolumeWork
     can [:create, :read, :edit, :update, :publish, :download], FileSet
     can [:create, :read, :edit, :update, :publish], Collection
-
+    can [:manage], ScannedResource
```
escowles: I don't think we want this — I think image editors shouldn't be allowed to delete ScannedResources that have ARKs assigned, for example.

app/models/ability.rb
```diff
@@ -89,7 +88,8 @@ def curation_concern_read_permissions
   end
 
   def readable_concern?(curation_concern)
-    !unreadable_states.include?(curation_concern.state)
+    # byebug
```
escowles: Please remove.

app/models/concerns/state_behavior.rb
```diff
@@ -2,38 +2,11 @@ module StateBehavior
...
-    validates_with StateValidator
-
-    def check_state
-      return unless state_changed?
-      complete_record if state == 'complete'
-      ReviewerMailer.notify(id, state).deliver_later
+    # Sipity workflow state name
+    def workflow_state
+      to_sipity_entity.workflow_state.name
+    rescue RuntimeError
```
escowles: Do we want to log this? Or is this just something that happens when things are in an inconsistent state, in tests, or something like that?

app/models/search_builder.rb
```diff
@@ -25,18 +24,27 @@ def join_from_parent(solr_params)
...
   def readable_states
-    StateWorkflow.aasm.states.map(&:name).map(&:to_s) - unreadable_states
+    all_states - unreadable_states
+  end
+
+  def all_states
+    Sipity::Workflow.all.map { |w| w.workflow_states.map(&:name) }.flatten.uniq
   end
 
+  delegate :unreadable_states, to: :current_ability
```
escowles: Do we need to move this down here? I'd rather keep delegations at the top of the class if we can.

config/workflows/geo_workflow.json
```diff
@@ -0,0 +1,127 @@
+{
+  "workflows": [
+    {
+      "name": "geo_works",
```
escowles: We should create a ticket to decide whether we need different book and geo workflows.  Having two workflows that differ only by their name/labels is a good placeholder in case we decide to change one later.  But if we really do want to use the same workflow, it would be nice to remove the duplication here.

escowles:
I'm seeing 8 errors when I run eslint that don't show up in Circle:

```
/Users/escowles/src/plum/app/assets/javascripts/flash.es6
  1:2  error  Parsing error: Illegal export declaration

/Users/escowles/src/plum/app/assets/javascripts/modal_viewer.es6
  1:2  error  Parsing error: Illegal export declaration

/Users/escowles/src/plum/app/assets/javascripts/plum_boot.es6
  1:2  error  Parsing error: Illegal import declaration

/Users/escowles/src/plum/app/assets/javascripts/server_uploader.es6
  1:2  error  Parsing error: Illegal export declaration

/Users/escowles/src/plum/app/assets/javascripts/shift_selecting.es6
  1:2  error  Parsing error: Illegal export declaration

/Users/escowles/src/plum/app/assets/javascripts/structure_manager.es6
  1:2  error  Parsing error: Illegal import declaration

/Users/escowles/src/plum/app/assets/javascripts/structure_parser.es6
  1:2  error  Parsing error: Illegal export declaration

/Users/escowles/src/plum/app/assets/javascripts/universal_viewer.es6
  1:2  error  Parsing error: Illegal export declaration

✖ 8 problems (8 errors, 0 warnings)
```

I'm not sure if that's related to the javascript test failures, or if I need to update eslint, or something else.

app/models/search_builder.rb
eliotjordan: That was an accident. Will move it back.

app/models/ability.rb
eliotjordan: Accidental addition. I'll remove it. I was poking at the abilities because of some permissions issues in the tests.

app/models/ability.rb
eliotjordan: Yep. Usually rubocop gets those.

app/models/concerns/state_behavior.rb
eliotjordan: Right. This is mainly about catching issues that arise during testing. Not all tests require full sipity workflows to be created, but this method gets called and throws an error.

spec/views/curation_concerns/base/_attributes.html.erb_spec.rb
```diff
@@ -24,6 +23,7 @@
   let(:can_edit) { false }
 
   before do
+    # byebug
     allow(view).to receive(:dom_class) { '' }
```
eliotjordan: remove

spec/views/curation_concerns/base/_attributes.html.erb_spec.rb
```diff
@@ -24,6 +23,7 @@
   let(:can_edit) { false }
 
   before do
+    # byebug
```
eliotjordan: remove

config/workflows/geo_workflow.json
eliotjordan: https://github.com/pulibrary/plum/issues/949

spec/values/cicognara_csv_spec.rb
```diff
@@ -13,12 +13,14 @@
...
   describe "#values", vcr: { cassette_name: "cicognara" } do
-    let(:obj) do
-      obj = FactoryGirl.create :scanned_resource_in_collection, source_metadata_identifier: '2068747'
+    let!(:obj) do
+      obj = FactoryGirl.create :complete_scanned_resource_in_collection, source_metadata_identifier: '2068747'
       obj.apply_remote_metadata
       obj.save!
       obj
     end
+    before do
```
escowles: Is this a remnant of something that was removed? Or does the empty block trigger something?  If it does, a comment would be good to make it clear.

eliotjordan: Remnant.

jpstroop:
Wow, hadn't noticed this was green. :tada:

app/models/concerns/state_behavior.rb
escowles: How about this: https://github.com/pulibrary/plum/pull/952

I'd feel better checking the workflow state carefully to avoid the errors than swallowing them, just in case there are other errors that could come from that.

PR #943, opened 2017-01-10
Refactoring welcome page to display blocks only when a user can create a work in that block

### LINKED ISSUE #926: Home page shows empty sessions if not logged in
If you aren't logged in, the home page has big empty sections where the buttons to add various types of works would go.

escowles:
Fixes #926


jpstroop:
Did this just slip through the cracks? Is it ready?

escowles:
Yes, it's ready and I think it just got overlooked.

PR #956, opened 2017-01-13
Adding rake task to migrate state property to Sipity

### LINKED ISSUE #955: Rake task to migrate state to Sipity

escowles:
Fixes #955

### DIFF
--- a/lib/tasks/state.rake
+++ b/lib/tasks/state.rake
@@ -1,0 +1,9 @@
+namespace :state do
+  desc "Migrate state from the state property to Sipity"
+  task migrate: :environment do
+    (MultiVolumeWork.all && ScannedResource.all).each do |r|
+      puts r.id
+      Workflow::InitializeState.call(r, 'book_works', r.state) unless r.workflow_state
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


eliotjordan:
The task should probably set the resource.state to `nil` or `Vocab::FedoraResourceStatus.active`. Otherwise, CC is going to hide them default.

tpendragon:
We should pull all the IDs from Solr and then `ActiveFedora::Base.find` each one, so we don't have to hold all those objects in memory.

lib/tasks/state.rake
```diff
@@ -0,0 +1,16 @@
+namespace :state do
+  desc "Migrate state from the state property to Sipity"
+  task migrate: :environment do
+    ['MultiVolumeWork', 'ScannedResource', 'ImageWork', 'RasterWork', 'VectorWork'].each do |model|
+      ActiveFedora::SolrService.get("has_model_ssim:#{model}")['response']['docs'].each do |doc|
```
tpendragon: Simpler/faster would be
```
ActiveFedora::SolrService.query("has_model_ssim:#{model}", fl: "id").map{|x| x["id"]}.each do |id|
```

escowles: 👍  — I also noticed that I had `book_works` hardcoded, so I replaced that with a hash of workflows to models they contain.

PR #973, opened 2017-01-23
Add pudl0135 and pudl0132 to pudl_collections.json

jpstroop:


jpstroop:
A couple of jobs can be retried after this

PR #976, opened 2017-01-24
Add pudl0117 to pudl_collections.json

jpstroop:


config/pudl_collections.json
```diff
@@ -274,6 +274,10 @@
     "slug": "pudl0132",
     "title": "Hizar-o Yek Shab",
     "blurb": "[WRITE ME]"
+  },
+  {
+    "slug": "pudl0117",
+    "title": "Attorney-General\'s Charges Against the Late Queen",
```
escowles: Do we need to escape this single quote?  Other single quotes are unescaped.

jpstroop: Oh, oops, maybe not.

PR #977, opened 2017-01-25
Upgrading to Rails 5

escowles:
* Also updating to ActiveFedora 11.1 and ActiveTriples 0.11

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'sass-rails', '~> 5.0'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '>= 1.3.0'
 # Use CoffeeScript for .coffee assets and views
-gem 'coffee-rails' #, '~> 4.1.0'
+gem 'coffee-rails', '~> 4.2.1'
 # See https://github.com/rails/execjs#readme for more supported runtimes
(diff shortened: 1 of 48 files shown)


escowles:
Now the tests pass, but both the jQuery uploader and the Browse Everything uploader have serious issues in the browser.  The jQuery uploader works, but fails to display the status of uploading items.  The Browse Everything uploader is completely broken.  No errors are logged in the Rails log or in the Javascript console in Chrome.

escowles:
I think the problem is that none of the ES6 is working, but I can't figure out why.  None of the relevant sprocket/babel gems have changed versions.  I downgraded browse_everything back to the same version in master, and still it doesn't work.

I was able to get it working with vanilla javascript (see https://github.com/pulibrary/plum/commit/785e86b9cc84c91a531482a6061de5ac2b4ba34a), so I'm pretty sure it's something with the ES6 and not the HTML, controller, etc.

tpendragon:
@escowles The problem was the `document.ready` in Blacklight wasn't working, so all of our `Blacklight.onLoad` code wasn't running. I'm not sure why that was the case, but a snippet in `application.js` fixes it.

jpstroop:
:tada:

PR #995, opened 2017-02-03
Changing Blacklight Javascript invocation

escowles:


escowles:
Or we could even just do:

```
$(document).ready(function(){
  Initializer = require('plum_boot')
  window.plum = new Initializer()
  Blacklight.activate()
})
```

Which works for me too.

PR #1050, opened 2017-02-27
Removing unnecessary/inappropriate instructions

### LINKED ISSUE #1042: Get rid of Batch upload link and the "the more metadata you provide..." text
Appears for SRs and MVWs, should check all.

screen shot 2017-02-27 at 11 31 43 am

escowles:
Fixes #1042


escowles:
The alternative to copying these views over is to update Hyrax to move all of these things out into separate partials that could be overridden.

tpendragon:
We can't i18n these away?

escowles:
One of them could be replaced with an i18n call we could override, but the other one includes a link, so the link would still be there even if the text was empty.    We could add a config option like the deposit agreement, since that's basically the same situation there...

PR #1062, opened 2017-03-01
Add rendering for ARK link

tpendragon:


tpendragon:
After this is merged + deployed we can make these ARKs be what's used by pomegranate.

PR #1091, opened 2017-03-14
Use jQuery 2 instead of 3

### LINKED ISSUE #1073: Facet "more" button is broken
Looks like a jQuery 3 size() problem.

tpendragon:
Blacklight's master branch is using deprecated #size() and we don't have the time to upgrade it at the moment, so in the meantime we can use jQuery 2 which hasn't removed the method yet.

Closes #1073


jpstroop:
Does this really not break anything else?

tpendragon:
Not that I noticed. We moved to 3 to fix some file manager dragging, but it seemed to work with the latest version of 2.

PR #1092, opened 2017-03-14
Looking up manifest by ARK

### LINKED ISSUE #965: Orangelight should be able to find a manifest for an item in Plum based on ARK
Orangelight has the ARK for an item, and needs to lookup the manifest URL

escowles:
Fixes #965

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -263,4 +263,17 @@ class CatalogController < ApplicationController
     return unless can? :create, ScannedResource
     blacklight_config.add_facet_field 'workflow_state_name_ssim', label: 'State'
   end
+
+  def lookup_manifest
+    ark = "#{params[:prefix]}/#{params[:naan]}/#{params[:arkid]}"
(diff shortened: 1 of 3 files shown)


jpstroop:
FWIW, I like this pattern more that what I had in mind. I was thinking it would be a route something like:

`/resolver?ark=ark:%2F%2F{naan}%2F{noid}`

which would give you a json body something like:

```json
{
  "ark" : "ark://{naan}/{noid}",
  "iiif" : "https://..."
}
```
but unless there are other keys we might want to include, better to just use HTTP as intended.

escowles:
👍  I think this will make it easy for OL to use: no need to encode the ARK, just append it to the base URL of the lookup service, and either get the manifest or a 404 back.

jpstroop:
I agree, and scoping the route to `/iiif` makes it clear. If we ever wanted to resolve PUDL identifiers, we could do `/pudl`, etc.

tpendragon:
Looks good. We'll have to see if UV can follow the redirect, but if it can't we can just parse it ourselves.

PR #1094, opened 2017-03-14
Rake task for bulk ingest

escowles:


tpendragon:
If this is something we're gonna support in any sort of ongoing fashion I'd like to extract the processing to a class and add some specs

escowles:
OK, I'll do that — I was thinking about refactoring the METS and PULFA background jobs to reduce the amount of duplicative ingest code too, so I might tackle that.

escowles:
TODO: pull out key book/geo configuration points into their own methods (see https://github.com/pulibrary/plum/issues/1101#issuecomment-287079408).

app/services/ingest_service.rb
```diff
@@ -0,0 +1,68 @@
...
+  def ingest_dir(dir, bib, user)
+    klass = choose_class(Dir["#{dir}/*"].first)
+    attribs = bib.nil? ? { title: [File.basename(dir)] } : { source_metadata_identifier: bib }
+    r = minimal_record klass, user, attribs
+    members = []
+    Dir["#{dir}/*"].sort.each do |f|
+      if File.directory? f
+        members << ingest_dir(f, nil, user)
+      else
+        members << ingest_file(r, f, user, file_set_attributes, title: [File.basename(f)])
```
eliotjordan: Looks like the `file_set_attributes` are being passed as file_options (visibility, lease, embargo) instead of attributes. Would something like this work? 
```
members << ingest_file(r, f, user, {}, file_set_attributes.merge({ title: [File.basename(f)] }))
```

escowles: Oops, you're right — I just pushed a fix for that.

eliotjordan:
This looks good to me. Still WIP?

escowles:
Nope, I think it's ready!

PR #1109, opened 2017-03-16
Updating PULFA metadata retrieval

escowles:
* Updating to latest pul_metadata_services
* Expanding test coverage of Pulfa metadata

### DIFF
--- a/app/jobs/ingest_pulfa_job.rb
+++ b/app/jobs/ingest_pulfa_job.rb
@@ -20,9 +20,8 @@ class IngestPULFAJob < ApplicationJob
   private
 
     def ingest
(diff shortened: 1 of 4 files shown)


app/jobs/ingest_pulfa_job.rb
```diff
@@ -40,6 +39,18 @@ def ingest
       r.save!
     end
 
+    def work_attributes
+      {
+        title: [@mets.xpath("//mets:structMap/mets:div/@LABEL").first.value],
```
jpstroop: This is only for the migration of the legacy content, right (because we won't have METS going forward)

escowles: Yes, this is the background job for ingesting the based on the legacy METS files.

PR #1110, opened 2017-03-17
Handle spaces in BrowseEverything paths.

tpendragon:


tpendragon:
So this throws an error on the LDP side for some reason..

PR #1114, opened 2017-03-20
Order browse everything ingests at the end

tpendragon:


app/services/browse_everything_ingester.rb
```diff
@@ -10,7 +10,7 @@ def initialize(curation_concern, upload_set_id, actor, file_info)
 
   def save
     actor.create_metadata({})
-    actor.attach_file_to_work(curation_concern)
+    # actor.attach_file_to_work(curation_concern)
```
jpstroop: rm?

PR #1139, opened 2017-03-24
Use Hyrax description property and add abstract property

### LINKED ISSUE #1132: Description is not imported from Voyager
The bibdata json-ld context maps the `description` term to `dc:description`:
https://bibdata.princeton.edu/context.json

This is the same predicate that Hyrax uses for the `description` property:
https://github.com/projecthydra-labs/hyrax/blob/master/app/models/concerns/hyrax/basic_metadata.rb#L24

However, in Plum this property is mapped to (dc terms?) `abstract`:
https://github.com/pulibrary/plum/blob/master/app/schemas/plum_schema.rb#L5

The result is that descriptions are not brought over when importing metadata from the bibdata service.

eliotjordan:
- Hyrax `description` property is now used instead of locally defined property.
- Adds `abstract` property.
- Removes description field from SR and MVW work forms.

Closes #1132

### DIFF
--- a/spec/services/polymorphic_manifest_builder_spec.rb
+++ b/spec/services/polymorphic_manifest_builder_spec.rb
@@ -280,7 +280,7 @@ RSpec.describe PolymorphicManifestBuilder, vcr: { cassette_name: "iiif_manifest"
       expect(result['@id']).to eq "http://plum.com/concern/scanned_resources/1/manifest"
     end
     it "has a description" do
-      expect(result.description).to eq [record.description]
+      expect(result.description).to eq Array(record.description)
     end
     context "when it has a bibdata ID" do
       it "links to seeAlso" do
--- a/app/adapters/remote_record.rb
+++ b/app/adapters/remote_record.rb
@@ -22,7 +22,7 @@ class RemoteRecord < SimpleDelegator
   def attributes
     result = super
     result[:date] = Array(result.delete(:date_created))
-    result[:description] = Array(result[:description]).first
+    result[:abstract] = Array(result[:abstract]).first
     result.delete :heldBy # TODO: map codes to locations (see plum#1001)
     result[:member_of_collections] = find_or_create(result.delete(:memberOf))
     result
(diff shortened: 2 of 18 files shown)


spec/services/polymorphic_manifest_builder_spec.rb
```diff
@@ -280,7 +280,7 @@ def build_file_set(id)
       expect(result['@id']).to eq "http://plum.com/concern/scanned_resources/1/manifest"
     end
     it "has a description" do
-      expect(result.description).to eq [record.description]
+      expect(result.description).to eq [record.description.first]
```
tampakis: This should be `expect(result.description).to eq record.description` right?

app/values/scanned_resource_pdf/renderer.rb
```diff
@@ -67,7 +67,7 @@ def metadata_hash
             hsh[entry["label"].to_sym] = Array(entry["value"]).join(", ")
           end
         result[:Title] = Array(scanned_resource.title).join(", ") if scanned_resource.title
-        result[:Description] = Array(scanned_resource.description).join(", ") if scanned_resource.description
+        result[:Description] = scanned_resource.description.join(", ") if scanned_resource.description
```
tampakis: I'm not sure this change is necessary. `Array([1,2,3])` leaves the array unchanged, while ensuring the description field returns an array.

spec/services/polymorphic_manifest_builder_spec.rb
eliotjordan: They aren't equivalent. `result.description` is and Array. `record.description` is and RDF something something class. The test fails unless you cast that into an Array.

eliotjordan: Updated to use `Array()` instead.

app/values/scanned_resource_pdf/renderer.rb
eliotjordan: Fair point. I'll revert the change.

tpendragon:
I'm :+1:, but I'll leave it to @tampakis.

PR #1157, opened 2017-04-07
allow geo filesets to be editable in UI

### LINKED ISSUE #1134: Can't edit geo works file sets
Raises an error:

`Missing partial hyrax/file_sets/_form_permission_under_lease,`

tampakis:
Closes #1134.
Hyrax uses simple_form, which requires these initializers: https://github.com/projecthydra-labs/hyrax/tree/1e504c200fd9c39120f514ac33cd42cd843de9fa/lib/generators/hyrax/templates/config/initializers
Added missing form partial.
Redirect geo works to fileset show page rather than file manager since geoworks doesn't use it.


tpendragon:
I'm surprised the initializers was an issue - I thought hydra-editor had all that stuff as part of it

eliotjordan:
I was able to get this to work without the `simple_form` initializer, but I did need the `simple_form_bootstrap` initializer. 

A few other things:
- The versions tab on the edit page should be removed since we aren't versioning files.
- The "Share With" section on the permissions tab should be removed.
screenshot 2017-04-10 12 25 46

app/views/hyrax/base/_form_permission.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <%= visibility_badge(Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED) %>
+      </label>
+      <label class="radio">
+        <%= f.radio_button :visibility, Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_EMBARGO %>
+        <%= render "form_permission_embargo", f: f %>
+      </label>
+      <label class="radio">
+        <%= f.radio_button :visibility, Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_LEASE %>
+        <%= render "form_permission_lease", f: f %>
+      </label>
```
eliotjordan: There is a bug (I think) in Hyrax where it applies a lease visibility to a file set if any of the lease parameters are not null; regardless of the value of the visibility param. The default lease dates in this form trigger the behavior, so file sets end up always having a lease. When I get a moment, I'll work up a PR to Hyrax about this. In the meantime, since we aren't using them in Plum yet, I think we can remove the lease and embargo sections here. Doing so fixes the issue.

PR #1172, opened 2017-04-12
Only create admin set when needed.

tpendragon:
Other specs were hitting Fedora whether they needed to or not, including
cleaning up, which was adding over a second per spec.


escowles:
This reduced my local test suite time by almost 50%! 🎉

PR #1173, opened 2017-04-12
Rake task to import vocabularies from CSV files

### LINKED ISSUE #1171: Rake task to load controlled vocabs from CSV files

escowles:
Fixes #1171


escowles:
This needs to be reworked now that we don't have VocabularyCollections

PR #1178, opened 2017-04-12
Removing the last traces of the 'Aardvark' name

escowles:


tpendragon:
Won't this break any existing jobs? Is the queue empty?

escowles:
There are 1.5M items in the `lowest` queue, so I don't think we'll have completely empty queues any time soon.  If changing the session name will break the background jobs, then we won't be able to do this any time soon 😞

tpendragon:
Yeah. Global ID generates the key using the application name. :(

PR #1196, opened 2017-04-19
Fix geoblacklight wms and iiif references for scanned maps

### LINKED ISSUE #1195: Geoblacklight references are incorrect for scanned maps
Two main issues:

- WMS/WFS references are being generated for scanned maps. This breaks proper display in pulmap.
- IIIF references are generated for restricted scanned maps in some cases.

eliotjordan:
Closes #1195

### DIFF
--- a/app/services/discovery/iiif_builder.rb
+++ b/app/services/discovery/iiif_builder.rb
@@ -36,6 +36,8 @@ module Discovery
       # @return [Bool]
       def geo_file_set?
         return false unless file_set
+        geo_mime_type = file_set.solr_document.fetch(:geo_mime_type_ssim, []).first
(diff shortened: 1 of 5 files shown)


app/services/discovery/iiif_builder.rb
```diff
@@ -40,6 +40,13 @@ def geo_file_set?
         @file_set_ids.include? file_set.id
       end
 
+      # Tests if the file set is a geo_works image format.
+      # @return [Bool]
+      def file_set_format?
```
escowles: Maybe roll this into `geo_file_set?` ?

eliotjordan: Rolled.

PR #1204, opened 2017-04-21
Initial pass of Box/Folder models + behavior

tpendragon:
There's a lot to do here still, but it's a good starting point so we can
divide and conquer.

### DIFF
--- a/app/models/ephemera_folder.rb
+++ b/app/models/ephemera_folder.rb
@@ -1,0 +1,31 @@
+# Generated via
+#  `rails generate hyrax:work EphemeraFolder`
+class EphemeraFolder < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::Hyrax::BasicMetadata
+  include ::StateBehavior
+  apply_schema EphemeraSchema, ActiveFedora::SchemaIndexingStrategy.new(
+    ActiveFedora::Indexers::GlobalIndexer.new([:symbol, :stored_searchable, :facetable])
+  )
+  # Change this to restrict which works can be added as a child.
+  # self.valid_child_concerns = []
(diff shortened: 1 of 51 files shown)


app/models/ephemera_folder.rb
```diff
@@ -0,0 +1,31 @@
...
+#  `rails generate hyrax:work EphemeraFolder`
+class EphemeraFolder < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::Hyrax::BasicMetadata
+  include ::StateBehavior
+  apply_schema EphemeraSchema, ActiveFedora::SchemaIndexingStrategy.new(
+    ActiveFedora::Indexers::GlobalIndexer.new([:symbol, :stored_searchable, :facetable])
+  )
+  # Change this to restrict which works can be added as a child.
+  # self.valid_child_concerns = []
```
escowles: Do we want to limit to FileSet?

app/services/workflow/updated_event.rb
```diff
@@ -1,6 +1,7 @@
 module Workflow
   class UpdatedEvent
     def self.call(target:, **)
+      return unless target.is_a?(ScannedResource) || target.is_a?(MultiVolumeWork)
```
escowles: Do we want to maintain the list of models here, or add something to the models that should generate events?

escowles:
I created #1205 and #1206 for some related issues that we can resolve once this is merged.

app/models/ephemera_folder.rb
tpendragon: Nah. This was meant to populate the "Add Child Work" drop down.

escowles:
I just pushed a commit that _should_ get us to 💯  — and then I think this is ready to merge.

PR #1211, opened 2017-04-21
Updating predicates for Ephemera properties

### LINKED ISSUE #1205: Boxes: Better Predicates
We should find better predicates (or add custom predicates to PULTerms) for the properties defined in: app/models/ephemera_box.rb and app/schemas/ephemera_schema.rb

escowles:
Fixes #1205

### DIFF
--- a/app/models/ephemera_box.rb
+++ b/app/models/ephemera_box.rb
@@ -7,9 +7,8 @@ class EphemeraBox < ActiveFedora::Base
   # Change this to restrict which works can be added as a child.
   self.valid_child_concerns = [EphemeraFolder]
   validates :title, presence: { message: 'Your work must have a title.' }
-  validates :box_number, :barcode, presence: true
-  validates :barcode, with: :barcode_valid?
-  property :barcode, predicate: ::RDF::Vocab::DC.identifier
+  validates :box_number, :identifier, presence: true
+  validates :identifier, with: :barcode_valid?
   property :box_number, predicate: ::RDF::RDFS.label
 
   self.human_readable_type = 'Ephemera Box'
@@ -20,7 +19,7 @@ class EphemeraBox < ActiveFedora::Base
   end
 
   def barcode_valid?
-    return true if Barcode.new(barcode.first).valid?
-    errors.add(:barcode, "has an invalid checkdigit")
+    return true if Barcode.new(identifier.first).valid?
+    errors.add(:identifier, "has an invalid checkdigit")
   end
 end
--- a/app/models/ephemera_box.rb
+++ b/app/models/ephemera_box.rb
@@ -9,8 +9,8 @@ class EphemeraBox < ActiveFedora::Base
   validates :title, presence: { message: 'Your work must have a title.' }
   validates :box_number, :barcode, presence: true
   validates :barcode, with: :barcode_valid?
-  property :barcode, predicate: ::RDF::URI("http://plum.com/predicates/barCode")
-  property :box_number, predicate: ::RDF::URI("http://plum.com/predicates/box_number")
+  property :barcode, predicate: ::RDF::Vocab::DC.identifier
+  property :box_number, predicate: ::RDF::RDFS.label
 
   self.human_readable_type = 'Ephemera Box'
   def box_number=(title)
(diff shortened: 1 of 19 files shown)


app/models/ephemera_box.rb
```diff
@@ -9,8 +9,8 @@ class EphemeraBox < ActiveFedora::Base
   validates :title, presence: { message: 'Your work must have a title.' }
   validates :box_number, :barcode, presence: true
   validates :barcode, with: :barcode_valid?
-  property :barcode, predicate: ::RDF::URI("http://plum.com/predicates/barCode")
-  property :box_number, predicate: ::RDF::URI("http://plum.com/predicates/box_number")
+  property :barcode, predicate: ::RDF::Vocab::DC.identifier
+  property :box_number, predicate: ::RDF::RDFS.label
```
escowles: Really not sure about this one — rdfs:label seems overly broad for this, but I wasn't able to find anything better.

app/schemas/ephemera_schema.rb
```diff
@@ -1,22 +1,21 @@
...
   # Language/Title from BasicMetadata
-  property :sort_title, predicate: ::RDF::URI("http://plum.com/predicates/sort_title")
-  property :alternative_title, predicate: ::RDF::URI("http://plum.com/predicates/alternative_title")
-  property :series, predicate: ::RDF::URI("http://plum.com/predicates/series")
+  property :sort_title, predicate: ::OpaqueMods.titleForSort
+  property :alternative_title, predicate: ::RDF::Vocab::DC.alternative
+  property :series, predicate: ::PULTerms.seriesTitle
   # Creator/Contributor/Publisher from BasicMetadata
-  property :geographic_origin, predicate: ::RDF::URI("http://plum.com/predicates/geo_origin")
+  property :geographic_origin, predicate: ::RDF::Vocab::BF2.originPlace
```
escowles: This expects a resource, so this may depend on how we implement the vocabularies

tpendragon: Should be fine. At the bare minimum we can make them URIs if we broadcast the RDF.

app/models/ephemera_box.rb
```diff
@@ -9,8 +9,8 @@ class EphemeraBox < ActiveFedora::Base
   validates :title, presence: { message: 'Your work must have a title.' }
   validates :box_number, :barcode, presence: true
   validates :barcode, with: :barcode_valid?
-  property :barcode, predicate: ::RDF::URI("http://plum.com/predicates/barCode")
-  property :box_number, predicate: ::RDF::URI("http://plum.com/predicates/box_number")
+  property :barcode, predicate: ::RDF::Vocab::DC.identifier
```
tpendragon: Doesn't this conflict with BasicMetadata's `:identifier` declaration?

escowles: Yes — maybe we should remove the property and just use identifier instead?

tpendragon: We'll want to update the label on the display, but otherwise that seems fine.

app/renderers/barcode_attribute_renderer.rb
```diff
@@ -0,0 +1,5 @@
+class BarcodeAttributeRenderer < Hyrax::Renderers::AttributeRenderer
+  def label
```
tpendragon: That's funny, I expected I18n to take care of this

escowles: It did for the forms, but I couldn't get the show view to work with i18n.

tpendragon: We'll have to figure that out later. Hyrax has spanish labels now for properties, so I wonder if it's our overrides causing problems.

escowles: Let me see if I can figure out how to get i18n to work from the Hyrax's Spanish translation file.

escowles: The issue is that Hyrax lets you have per-model i18n, but AFAICT, Blacklight doesn't.  I think it would be worth exploring whether that's possible, or at least using i18n in this class at least.

PR #1218, opened 2017-04-25
Add DataTables for Folder View

### LINKED ISSUE #1188: Add datatables sorting to list of folders in box#edit view
This was on their list for the 4/18 meeting, but we forgot to cover it. They would like to be able to quickly sort this table:

screen shot 2017-04-19 at 9 37 16 am

tpendragon:
Closes #1188

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -132,3 +132,4 @@ group :staging, :production do
   gem 'ddtrace'
 end
 gem 'honeybadger', '~> 2.0'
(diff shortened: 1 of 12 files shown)


tpendragon:
Coverage for this isn't gonna be done - to fix it I'm going to have to get EphemeraFolderController to inherit from `HyraxController`, which opens up a big can of worms with expected schema properties and a conflict with single-valued property behavior. So, need to spin on this a bit. May have a supporting PR for redoing single-valued properties first.

PR #1223, opened 2017-04-27
Remove Multiple: False

tpendragon:
It's gotten hard to handle which fields are allowed to be multi-valued
and which aren't. In order to deal with conflicts and make the reasoning
much easier to handle, make everything multi-valued and handle
single-valued display via the forms (SingleValuedForm).

### DIFF
--- a/app/forms/hyrax/hyrax_form.rb
+++ b/app/forms/hyrax/hyrax_form.rb
@@ -1,7 +1,9 @@
 module Hyrax
-  class HyraxForm < Hyrax::Forms::WorkForm
+  class HyraxForm < ::Hyrax::Forms::WorkForm
+    include SingleValuedForm
     self.terms += [:holding_location, :rights_statement, :rights_note, :source_metadata_identifier, :portion_note, :description, :abstract, :state, :member_of_collection_ids, :ocr_language, :nav_date, :pdf_type, :start_canvas, :uploaded_files]
     self.required_fields = [:title, :source_metadata_identifier, :rights_statement]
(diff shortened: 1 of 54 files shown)


app/forms/hyrax/hyrax_form.rb
```diff
@@ -11,8 +13,6 @@ def notable_rights_statement?
     def self.multiple?(field)
```
escowles: Do we need to keep overriding `multiple?` here?

app/forms/hyrax/hyrax_form.rb
```diff
@@ -46,7 +46,7 @@ def rights_statement
       if Array(self["rights_statement"]).first.blank?
         "http://rightsstatements.org/vocab/NKC/1.0/"
       else
-        self["rights_statement"]
+        Array(self["rights_statement"]).first
```
escowles: I think this could be simplified to something like:

```
self['rights_statement'] || 'http://rightsstatements.org/vocab/NKC/1.0/'
```

escowles:
Looks good — there's just a couple of minor things that could be nice to clean up.

app/forms/hyrax/hyrax_form.rb
tpendragon: I think it's only necessary for `identifier`, which needs weird behavior because of expectations in Grocer (we should fix that, but I didn't wanna cross gem boundaries for this)

escowles: OK, that's fine — I've created an issue in Grocer for this: https://github.com/pulibrary/grocer/issues/24

app/forms/hyrax/hyrax_form.rb
tpendragon: It looks like self['rights_statement'] is getting initialized to an empty string, but I can make the conditional do a bunch less casting I think.

PR #1234, opened 2017-05-01
hide description and subject fields from search results display

### LINKED ISSUE #1099: Search result list display options
Can we adjust the data which displays in a search result list in Plum?  There is a lot of data displaying now.  I would like to remove some of what is there  (subject, resource type, file format) and I would like to add in the collections the object is associated with and its status.

What I am trying to do now, for example, is find the 41 objects in the Yemeni collection which aren't also in the Islamic manuscript collection.  It would be easier if I could exclude the Islamic collection, but barring that, I could manage if I could see the collections listed.

tampakis:
Closes #907.
Closes #1099.


escowles:
@tampakis We talked about this with Joyce yesterday, and updated #1099 with a much smaller list of fields that should be displayed in the search results.  Can you update this PR?

app/models/solr_document.rb
```diff
@@ -40,10 +40,6 @@ def logical_order
       end
   end
 
-  def language_display
```
tpendragon: This isn't necessary for facets? I don't remember why this was around..

tampakis: nope just for display with the individual record on the search results. The language gets mapped at index time for the facet: https://github.com/pulibrary/plum/blob/master/app/indexers/work_indexer.rb#L27-L29

PR #1273, opened 2017-05-09
Add Category & Subject CV field.

tpendragon:
This uses `<optgroup>` and bootstrap-select instead of the home-built
javascript picker in the current LAE. I hope this is sufficient, because
the HTML is much more standard, and even without bootstrap-select you
get a nice categorized multiple-select box.
screen shot 2017-05-09 at 3 25 04 pm

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -28,7 +28,7 @@ class WorkIndexer < Hyrax::WorkIndexer
         solr_doc[Solrizer.solr_name("#{field}_literals", :symbol)] = output
       end
       solr_doc[Solrizer.solr_name("identifier", :symbol)] = object.identifier
-      [:geo_subject, :geographic_origin, :genre].each do |property|
+      [:geo_subject, :geographic_origin, :genre, :subject].each do |property|
         next unless object.respond_to?(property)
         solr_doc[Solrizer.solr_name(property.to_s, :facetable)] = object.send(property).map do |code|
           authority = AuthorityFinder.for(property: property, model: object)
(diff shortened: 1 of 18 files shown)


jpstroop:
Have you tried this with all of the subjects loaded? My concern (if I understand the UI correctly from the screenshot), is that the drop-down might get very long.

escowles:
I worried about the same thing @jpstroop — the current LAE subject list has 440 terms in 18 groups.  But I loaded them locally, and it works fine.  It would be a long list to scroll through, but you can type a couple of letters from a subject and pick it quite easily.  So I think it's worth doing a demo to see what everybody thinks of it (and making sure to do a local demo, so the experience isn't obscured by the limits of screensharing).

tpendragon:
@jpstroop Yeah, that screenshot above is with all the subjects loaded. It relies reasonably heavily on the search aspect, so it depends on how often they actually know what subject they're after.

As for the demo, since it's all JS it should be just fine to demo remotely.

jpstroop:
👍  Let's see what they think. I am worries about how well they might or might not know the vocab, but agree this is a much nicer solution if we can get away with it.

app/indexers/work_indexer.rb
```diff
@@ -28,7 +28,7 @@ def generate_solr_document
         solr_doc[Solrizer.solr_name("#{field}_literals", :symbol)] = output
       end
       solr_doc[Solrizer.solr_name("identifier", :symbol)] = object.identifier
-      [:geo_subject, :geographic_origin, :genre].each do |property|
+      [:geo_subject, :geographic_origin, :genre, :subject].each do |property|
```
escowles: Do we want to index the subject category like the current LAE app does?

escowles:
One other small note: I had to update my solr_wrapper config to use `instance_directory` instead of `instance_dir` — I'll open a separate PR to update solr_wrapper and fix that, though.

PR #1285, opened 2017-05-11
Fixes specs for cleanup-derivatives method

eliotjordan:
The specs for `cleanup-derivatives` in #1282 weren't actually testing if the derivative was removed or not. The main reason is the `derivatives_for_reference` should return an array rather than a single value.


eliotjordan:
@escowles I started reviewing #1282 this morning, but I got caught up in meetings and didn't get my comments in before it was merged. Sorry about that.

escowles:
Thanks for catching this @eliotjordan — I completely spaced on that, putting `true` instead of `false` and not noticing that the test wasn't really testing what it was supposed to.

PR #1287, opened 2017-05-13
Task, job, and service for ingesting MapSets

### LINKED ISSUE #1259: Create tasks for ingesting for sets of scanned maps

eliotjordan:
Closes #1259

### DIFF
--- a/app/services/ingest_map_set_service.rb
+++ b/app/services/ingest_map_set_service.rb
@@ -1,0 +1,105 @@
+class IngestMapSetService < IngestScannedMapsService
+  attr_reader :record, :base_file_path, :user
+
+  def ingest_map_set(record, base_file_path, user)
+    @record = record
+    @base_file_path = base_file_path
+    @user = user
+
+    delete_duplicates!("source_metadata_identifier_tesim:\"#{RSolr.solr_escape(record['bibid'])}\"") if record['bibid']
+    map_set = create_map_set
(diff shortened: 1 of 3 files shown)


app/services/ingest_map_set_service.rb
```diff
@@ -0,0 +1,105 @@
...
+  def workflow_state(klass)
+    klass == MapSet ? 'final_review' : 'complete'
+  end
+
+  def ingest_work(file_path, ark, title, sheet_title)
+    klass = ImageWork
+    bib_id = record['bibid']
+    attribs = { source_metadata_identifier: [bib_id], identifier: ["ark:/88435/#{ark}"] }
+    r = minimal_record klass, user, attribs
+    r.source_metadata_identifier = []
```
tampakis: Why do we remove the source metadata identifier from the child image works?

tampakis:
Looks great. Just one question left in a code line comment.

app/services/ingest_map_set_service.rb
eliotjordan: Because it's not really the source metadata for that record, we're really just pulling from the parent MapSet and then changing the title. And bounding box (one day). If someone refreshed all the image works from bibdata, it would overwrite the correct title.

tampakis: 👍

PR #1290, opened 2017-05-16
Add workflow for LAE Folders

tpendragon:
1. Adds a two-step workflow for Ephemera Folders. Immediately upon
ingest the metadata record is public, but the manifest is locked down.

2. When QA of images is complete, the manifest is made public.

3. Ephemera Editors can back out the workflow from Complete -> Needs QA.

This PR also fixes attribute display of rights/state for Ephemera
Folders.

### DIFF
--- a/app/services/workflow/plum_workflow_strategy.rb
+++ b/app/services/workflow/plum_workflow_strategy.rb
@@ -9,6 +9,7 @@ module Workflow
     # @return [String] The name of the workflow to use
     def workflow
       return Sipity::Workflow.where(name: 'book_works').first! if book_works.include? work_class
+      return Sipity::Workflow.where(name: 'folder_works').first! if folder_works.include? work_class
       return Sipity::Workflow.where(name: 'geo_works').first! if geo_works.include? work_class
     end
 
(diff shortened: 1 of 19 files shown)


spec/ability/roles_spec.rb
```diff
@@ -44,6 +44,14 @@
     FactoryGirl.create(:flagged_scanned_resource, user: image_editor, identifier: ['ark:/99999/fk4445wg45'])
   }
 
+  let(:complete_ephemera_folder) {
```
escowles: I don't see any assertions about this object — do we need it?  Or do you want to add some assertions about being able to access a complete object?

app/services/workflow/plum_workflow_strategy.rb
```diff
@@ -19,7 +20,11 @@ def work_class
       end
 
       def book_works
-        %w(ScannedResource MultiVolumeWork EphemeraBox EphemeraFolder)
+        %w(ScannedResource MultiVolumeWork EphemeraBox)
+      end
+
+      def folder_works
```
escowles: I expected this to be called ephemera_works — but maybe there are other kinds of things with a similar workflow/organization.

tpendragon: I named it this because Boxes will probably have a different workflow (due to shipping)

spec/ability/roles_spec.rb
tpendragon: Yes, I meant to check to make sure anonymous users could `manifest` them. Added it.

Issue #1302, opened 2017-05-19
New LAE role for editors


jpstroop:
* should not be able to delete
 * should not be able to manage vocabs

escowles:
We have an Ephemera Editor role, but these users can delete their own files and Ephemera Folder/Box objects.  Is this OK?

PR #1344, opened 2017-06-08
ActiveRecord-based configuration for mapping fields to Vocabularies

### LINKED ISSUE #1255: UI for linking fields to vocabs
Maybe each model could define the fields that could be controlled vocab fields, e.g. `self.controlled_fields = [:genre, :language, :subject]`

Then we could have an AR model that let you pick a Vocab and link it to a field like `EphemeraFolder.genre`.  There would need to be a helper to interrogate all of the CurationConcerns to see which ones had controllable fields for generating the forms.  But once those records were created, AuthorityFinder could use that data to select for forms, etc.

I'm not sure how this fits in with templates and having different profiles/templates/whatever for LAE vs. South Asian Ephemera, though.  Maybe the template could be a container for the Vocab/field mappings?

escowles:
Adds two new ActiveRecord models for managing Ephemera:
* EphemeraProject, which corresponds to a set of EphemeraBoxes/EphemeraFolders using the same controlled vocabularies.
* EphemeraField, which links a field to a controlled vocabulary in the context of a project.

The new workflow for creating a new ephemera project is:
1. Create a new EphemeraProject for the project
2. (Optional) Create a new EphemeraField for each field that should linked to Vocabulary
3. Create a EphemeraBox in the EphemeraProject
4. Add EphemeraFolders (where the form fields will use the Vocabularies selected in step 2).

Fixes #1255


tpendragon:
Spinning this up to take a look now

app/assets/javascripts/ephemera_fields.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
tpendragon: Delete?

app/assets/javascripts/ephemera_projects.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
tpendragon: Delete?

app/assets/stylesheets/ephemera_fields.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the EphemeraFields controller here.
```
tpendragon: Delete?

app/assets/stylesheets/ephemera_projects.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the EphemeraProjects controller here.
```
tpendragon: Delete?

escowles:
Should we index + facet "Ephemera Project"? Or maybe associate it with a collection?

I've deleted the unused javascripts/stylesheets and added some navigation to the front page.  I've also added basic access control enforcement, but will defer more auth work to #1302 

For indexing, the box does index the EphemeraProject name and id, and that displays on the show page.  EphemeraFolders just index the id so it can be used to find the appropriate controlled vocab fields.  We could also index the EphemeraProject name there too — maybe we should just index it the same way as Collections, so we could effectively use the EphemeraProject instead of Collections?

tpendragon:
> maybe we should just index it the same way as Collections, so we could effectively use the EphemeraProject instead of Collections

:+1:

escowles:
I was torn about this because I think the EphemeraProject is going to map to a Collection, but clearly serves a very different purpose.  And I definitely don't want to have a bunch of redundant Collection behavior on EphemeraProject.

So I'll index it as its own thing and provide a link from the EphemeraFolder like is already there for EphemeraBox, to make it easier to navigate up to the project.

PR #1350, opened 2017-06-09
Memoize messaging clients

### LINKED ISSUE #1349: Messaging events open multiple RabbitMQ channels
When messages passed to RabbitMQ, multiple connections and channels are opened. It looks as if one channel is opened per message. This quickly saturates Rabbit when sending messages in bulk.

eliotjordan:
I think this might solve the problem with multiple RabbitMQ connections. The messaging clients are initialized only once during startup. @tpendragon does this seem right to you?

Closes #1349

### DIFF
--- a/config/initializers/plum_config.rb
+++ b/config/initializers/plum_config.rb
@@ -4,11 +4,11 @@ module Plum
   end
 
   def messaging_client
-    MessagingClient.new(Plum.config['events']['server'])
+    @messaging_client ||= MessagingClient.new(Plum.config['events']['server'])
   end
 
   def geoblacklight_messaging_client
-    GeoblacklightMessagingClient.new(Plum.config['events']['server'])
+    @geoblacklight_messaging_client ||= GeoblacklightMessagingClient.new(Plum.config['events']['server'])
   end
 
   private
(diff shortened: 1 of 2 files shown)


eliotjordan:
Or... perhaps the connection should be closed after each publish event. Is this more a resilient approach?

```
  def publish(message)
    exchange.publish(message, persistent: true)
    bunny_client.stop
  rescue
    Rails.logger.warn "Unable to publish message to #{amqp_url}"
  end
```

IIRC Rails creates a controller instance for each request, so lines like this shouldn't be a problem: https://github.com/pulibrary/plum/blob/master/app/controllers/hyrax/hyrax_controller.rb#L64

tpendragon:
This looks good to me.

jpstroop:
Would it make sense to wrap the connection in a `do` block, so that you can
catch errors but always make sure the connection gets closed?

Look at how the Oracle connection is made in VoyagerHelpers.

-Js

jpstroop:
https://github.com/pulibrary/voyager_helpers/blob/master/lib/voyager_helpers/oracle_connection.rb

eliotjordan:
Reading up on RabbitMQ best practices, it seems that having a long running channel is better than opening and closing one for each publishing event. With that in mind, I'll keep the code as it is and work on fixing the failing specs.

eliotjordan:
@tpendragon Tested this out on `hydra-dev`. Works as promised. Number of open connections remains static regardless of the number of events sent from plum.

screenshot 2017-06-13 12 08 20

PR #1371, opened 2017-06-22
Caching and Performance Improvements

### LINKED ISSUE #1347: Caching on GeoBlacklight document
We might want to enable caching on the GeoBlacklight document, similar to the way IIIF manifests are cached. As it stands, reindexing GeoBlacklight takes several days when generating each doc from scratch.

eliotjordan:
- Caches geoblacklight documents.
- Stores `default_url_options` in Plum config. Profiling shows that calls to `ActionMailer::Base.default_url_options` are the single slowest step in generating documents.
- Uses GeoWorks 0.1.3 which includes several performance enhancements.

Closes #1347


eliotjordan:
@tpendragon @escowles You, might want to take a look at this because it touches non-geo Plum in several places.

escowles:
👍

PR #1386, opened 2017-06-30
Update deploy script for Xenial worker boxes.

tpendragon:


config/deploy/production.rb
```diff
@@ -9,7 +9,7 @@
 server 'lib-plum1', user: 'deploy', roles: %w(app db web)
 server 'lib-proc4', user: 'deploy', roles: %w(worker worker2)
 server 'lib-proc5', user: 'deploy', roles: %w(worker worker2)
-server 'lib-proc6', user: 'deploy', roles: %w(worker3)
+server 'lib-proc6', user: 'deploy', roles: %w(worker2)
```
escowles: Do we want to make them all the same, or keep the derivatives / ingest+derivatives split?

tpendragon: Right now it's this way because I haven't run the ansible scripts on lib-proc6.

PR #1387, opened 2017-06-30
Adding Capistrano command to reload workflows

### LINKED ISSUE #1385: Create capistrano task to reload workflows
We need to do this when we update our workflows, but not on every deploy

escowles:
Fixes #1385

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -47,6 +47,15 @@ namespace :deploy do
       end
     end
   end
+  task :workflow do
+    on roles(:web) do
+      within release_path do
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -47,6 +47,15 @@
       end
     end
   end
+  task :workflow do
```
tpendragon: Should probably just be on the web role. Don't wanna run it more than once.

escowles: Good point — updated to just run on the web role.

PR #1403, opened 2017-07-12
[WIP] Overrides the DateBuilder Service for GeoWorks::Discovery::DocumentBuilder

jrgriffiniii:
Resolves #1401 by ensuring that layer years are parsed from the "date" attribute when "temporal" is empty

PR #1404, opened 2017-07-12
Overrides the DateBuilder Service for GeoWorks DocumentBuilder instances

### LINKED ISSUE #1401: Make sure publication dates get indexed for geoworks
Pull from date field in addition to temporal. Override GeoWorks method.

https://github.com/samvera-labs/geo_works/blob/master/app/services/geo_works/discovery/document_builder/date_builder.rb#L25

jrgriffiniii:
Resolves #1401 by ensuring that layer years are parsed from the "date" attribute when "temporal" is empty.  Please note that this requires that `date` be exposed by geospatial Presenters (see #1403).

### DIFF
--- a/spec/services/discovery/date_builder_spec.rb
+++ b/spec/services/discovery/date_builder_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+
+describe Discovery::DateBuilder do
+  subject { described_class.new(geo_work_presenter) }
+
+  let(:document) { instance_double('Document') }
+  let(:geo_work_presenter) { VectorWorkShowPresenter.new(SolrDocument.new(geo_work.to_solr), nil) }
+
+  describe 'geospatial work document' do
+    before do
+      allow(document).to receive(:layer_modified=)
+      allow(document).to receive(:issued=)
+    end
+
+    context 'with temporal values' do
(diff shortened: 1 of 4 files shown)


spec/services/discovery/date_builder_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+describe Discovery::DateBuilder do
+  subject { described_class.new(geo_work_presenter) }
+
+  let(:document) { instance_double('Document') }
+  let(:geo_work_presenter) { VectorWorkShowPresenter.new(SolrDocument.new(geo_work.to_solr), nil) }
+
+  describe 'geospatial work document' do
+    before do
+      expect(document).to receive(:layer_modified=)
+      expect(document).to receive(:issued=)
```
eliotjordan: I'd use `allow` instead of `expect` in the before block. We're stubbing those methods rather than testing if the code calls them.

eliotjordan:
One last change. We need to include this module in the document builder services initializer. If you replace this line: https://github.com/pulibrary/plum/blob/master/config/initializers/plum_config.rb#L34
with `Discovery::DateBuilder`, GeoWorks will use the new `layer_year` method when generating the gbl document.

PR #1423, opened 2017-07-26
[WIP] Setting read_groups in bulk ingest so gated discovery works as expected

### LINKED ISSUE #1419: Permissions enforcement not working
Possibly only in bulk-ingested items, permissions are not being enforced correctly.  In particular, users who should be able to view any item (e.g. fulfillers) are not able to find or view some items.

escowles:
Fixes #1419


escowles:
I'm going to deploy this to staging and make sure it works as expected.

escowles:
After setting read_groups to `['public']` if it was empty, I can see all of the ScannedResources in staging whether I'm an admin or a fulfiller.

escowles:
Closing in favor of solution that works for collections too.

PR #1456, opened 2017-08-16
Ingest volumes first, then combine into MultiVolumeWork

escowles:


escowles:
Sample work ingested with this branch: https://plum.princeton.edu/concern/multi_volume_works/p1z40pr89n

PR #1458, opened 2017-08-16
Ensures that the Universal Viewer is not rendered for invalid IIIF resources

### LINKED ISSUE #1429: Manifest with no sequences when no images are loaded
I'm getting an error with this manifest, it looks like no images are loaded and so no sequence gets generated (which isn't valid IIIF)
https://plum.princeton.edu/concern/scanned_resources/prr174n35m/manifest

Probably there should be images loaded, though?

jrgriffiniii:
Resolves #1429 by ensuring that the Universal Viewer is not rendered for resources where the IIIF Manifest is empty or invalid

### DIFF
--- a/spec/controllers/hyrax/collections_controller_spec.rb
+++ b/spec/controllers/hyrax/collections_controller_spec.rb
@@ -3,17 +3,18 @@ require 'rails_helper'
 RSpec.describe Hyrax::CollectionsController do
   describe "#manifest" do
     let(:user) { FactoryGirl.create(:admin) }
+
     before do
       sign_in user if user
     end
+
     it "returns a manifest for the collection" do
       coll = FactoryGirl.create(:collection)
-      allow(ManifestBuilder).to receive(:new).and_call_original
 
       get :manifest, params: { id: coll.id, format: :json }
 
-      expect(ManifestBuilder).to have_received(:new)
-      expect(response).to be_success
+      expect(response.status).to eq 200
+      expect(response.body).not_to be_empty
     end
 
     context 'when the manifest is empty or invalid' do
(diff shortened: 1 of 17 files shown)


spec/controllers/hyrax/collections_controller_spec.rb
```diff
@@ -16,6 +16,35 @@
...
+      end
+
+      it 'returns an error message' do
+        allow(manifest).to receive(:to_json).and_return("{}")
+        allow(ManifestBuilder).to receive(:new).and_return(manifest)
+
+        get :manifest, params: { id: coll.id, format: :json }
+
+        response_json = JSON.parse(response.body)
+        expect(response_json).to be_empty
```
escowles: I think it would be good to also check the response status code.

spec/controllers/hyrax/ephemera_folders_controller_spec.rb
```diff
@@ -109,6 +109,30 @@
...
+          expect(EphemeraFolder).not_to receive(:find)
+
+          allow(manifest).to receive(:to_json).and_return("{}")
+          allow(ManifestBuilder).to receive(:new).and_return(manifest)
+          expect(Rails.logger).to receive(:warn).with("Image cannot be viewed at this time.")
+
+          get :manifest, params: { id: "test2", format: :json }
+
+          response_json = JSON.parse(response.body)
+          expect(response_json).to be_empty
```
escowles: Same as above: we should check the status code, too.

escowles:
This looks good — the only changes I'd like to see are making sure we get the response status codes we expect, since they're important for triggering client error handling behavior.

PR #1467, opened 2017-08-24
Lae ingest

### LINKED ISSUE #1193: LAE Migration - No PDFs
Do not allow PDF download by default

escowles:
Fixes #1193


app/services/ingest_ephemera_service.rb
```diff
@@ -0,0 +1,156 @@
+require 'rdf/ntriples'
+require 'vocab/pul_store'
+
+class IngestEphemeraService
+  def initialize(folder_dir, user, project, logger = Logger.new(STDOUT))
+    @folder_dir = folder_dir
+    @user = user
+    @project = project
+    @logger = logger
+    @vocabs = {}
```
tpendragon: Can we extract everything not passed in to private methods?

tpendragon:
Just a small thing and we should be good.

app/services/ingest_ephemera_service.rb
escowles: 👍  — I made everything private except for the main `ingest` method and the `lookup` method that's tested separately.

